PR #1, opened 2022-04-05
Review and update ESLint rules

lukebennett88:
- Update ESLint packages
- Set up root level `.eslintrc.js` file (heavily inspired by [this one](https://github.com/keystonejs/keystone/blob/main/.eslintrc.js))
- Remove seperate `.eslintrc.js` file in Next app
- Update lint scripts to use `.gitignore` for ignore path

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,126 @@
+const OFF = 0;
+const WARN = 1;
+const ERROR = 2;
+
+const reactComponentTypeMessage = {
+  message:
+    'This type includes the children prop which is generally wrong, instead of using this type, type the props of the component',
+};
+
+/** @type {import('@types/eslint').Linter.BaseConfig} */
+module.exports = {
+  parser: '@typescript-eslint/parser',
+  env: {
+    browser: true,
+    es6: true,
+    node: true,
+  },
+  extends: ['next', 'prettier'],
+  plugins: [
+    'simple-import-sort',
+    'import',
+    '@typescript-eslint',
+    '@preconstruct/format-js-tag',
+  ],
+  settings: {
+    next: { rootDir: ['docs/', 'packages/*/'] },
+    react: { version: 'detect' },
+  },
+  rules: {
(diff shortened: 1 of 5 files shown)


lukebennett88:
Note: I haven't added any of the Jest rules yet.
Potentially we might use [Vitest](https://vitest.dev/) instead?
Worth discussing before we implement.

.eslintrc.js
```diff
@@ -0,0 +1,126 @@
...
+    {
+      files: ['**/*.{ts,tsx}'],
+      rules: {
+        // TypeScript already checks for the following things and they conflict with TypeScript
+        'import/no-unresolved': OFF,
+        'no-undef': OFF,
+      },
+    },
+  ],
+};
```
Thinkscape, line 126: Where do those come from? The previous repo?

Thinkscape (approved):

.eslintrc.js
lukebennett88: > Set up root level .eslintrc.js file (heavily inspired by this one)
Based on Keystone's ESLint rules

package.json
```diff
@@ -9,8 +9,8 @@
     "format": "prettier --check .",
     "format:check": "prettier --check .",
     "format:write": "prettier --write .",
-    "lint:fix": "eslint --fix .",
-    "lint": "eslint .",
+    "lint:fix": "yarn run lint --fix",
+    "lint": "eslint . --ignore-path \"./.gitignore\"",
```
lukebennett88, line 13: Using our `.gitignore` for our ignore path. I haven't seen anyone else do this, so not sure if its a bad idea or I'm a genius, but having a `.gitignore` a `.eslintignore` and a `.prettierignore` all in the same directory, all with pretty much the same set of files has never really made a whole lot of sense to me...

Thinkscape: I vote for "genius" :D 
I wish we could call it `.ignore`... but we could put a comment inside it, to indicate that it's being used this way.
Cool idea!

oscargws (approved):

PR #4, opened 2022-04-07
Set up tests

nderkim:
Get tests working again


package.json
```diff
@@ -12,7 +12,8 @@
     "lint:fix": "eslint --fix .",
     "lint": "eslint .",
     "postinstall": "preconstruct dev && manypkg check",
-    "typecheck": "yarn postinstall && ultra -r tsc --noEmit"
+    "typecheck": "yarn postinstall && ultra -r tsc --noEmit",
+    "test": "jest"
```
Thinkscape, line 16: How about:
```
"test:unit": "jest",
"test": "yarn run test:unit && yarn run typecheck",
```

I like my `test` command to test all-the-things without me having to consider what it tests and what it doesn't.

nderkim: done ✅

package.json
```diff
@@ -12,7 +12,9 @@
     "lint:fix": "yarn run lint --fix",
     "lint": "eslint . --ignore-path \"./.gitignore\"",
     "postinstall": "preconstruct dev && manypkg check",
-    "typecheck": "yarn postinstall && ultra -r tsc --noEmit"
+    "typecheck": "tsc --noEmit",
```
lukebennett88, line 15: I think we should keep the `postinstall` here. If you were to run `yarn build` before running `yarn typecheck` you will get a bunch of false positive type errors.

nderkim: I think the type errors are valid errors. eg. `level` is declared as a `number` but is being passing in as a `string` throughout the codebase.
Screen Shot 2022-04-08 at 8 48 00 am

lukebennett88: If you run `preconstruct dev` it should be a number ¯\_(ツ)_/¯ :
PixelSnap 2022-04-08 at 08 57 07@2x

freemany (approved):

package.json
nderkim: @Thinkscape looks like there is only one `tsconfig` in this repo, so only one process to run

lukebennett88 (approved):

PR #6, opened 2022-04-07
Move fonts out of `@spark-web/core`

lukebennett88:


JedWatson:
We can backlog this if you want (so it becomes a separate task) but I think this needs to be easier for consumers to understand how to set up correctly.

It would be good to have a clearer link between

* Default font stack set in theme
* Font inclusions to add in the html document

PR #9, opened 2022-04-11
feat: added text area component with corresponding docs

justinfaynhan:


matildaPan (approved):
🎉

packages/text-area/src/TextArea.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const theme = useTheme();
+  const inputStyles = useInput({
+    disabled,
+    invalid,
+  });
+
+  return {
+    ...inputStyles,
+    overflow: 'hidden', // fix for Safari to prevent unwanted scrolling of parent container to occur
+    textOverflow: 'ellipsis',
```
lukebennett88, line 77: We can remove these, they were specific for the `Select`

packages/text-area/src/TextArea.tsx
```diff
@@ -0,0 +1,90 @@
...
+    resize: 'vertical',
+    minHeight: theme.sizing.medium,
+
+    // Prevent text going underneath the chevron icon
+    paddingRight:
+      theme.sizing.xxsmall + // size of chevron icon
+      theme.spacing.medium * 2, // paddingX value
```
lukebennett88, line 84: Same as above. This was reserving space for the chevron icon for the `Select` so we don't need it.

lukebennett88 (changes requested):

packages/text-area/README.md
```diff
@@ -0,0 +1,47 @@
...
+        <TextArea value={textInput} onChange={event => setTextInput(event.target.value)}/>
+      </Field>
+      <Text>You have inputed: </Text>
+      {textInput && (
+        <Text>{textInput}</Text>
+      )}
+    </Stack>
+  )
+```
+
```
lukebennett88, line 32: Happy to approve this as is, but feels a little lite. Would be nice to flesh this out a tiny bit more.

[Balance has a size prop](https://balance.reckon.com/package/textarea#size) @jordangeizer do you think we should do the same?

justinfaynhan: Added more doc examples regarding text area demonstrating size and weight use cases

lukebennett88 (approved):

PR #11, opened 2022-04-11
Add test coverage and checkbox tests

freemany:
Screen Shot 2022-04-11 at 6 21 20 pm


packages/checkbox/src/checkbox.tsx
```diff
@@ -21,7 +21,7 @@ export const Checkbox = forwardRef<HTMLInputElement, CheckboxProps>(
   ) => {
     const { inputId, messageId } = useFieldIds(idProp);
     const a11yProps = {
-      'aria-describedby': message && messageId,
+      'aria-describedby': message ?? messageId,
```
lukebennett88, line 24: This was correct before.
`aria-describedby` should be provided an id. We're saying if we have a message, the checkbox is described by that message, otherwise it should be `undefined`.

lukebennett88, line 24: This was correct before.
`aria-describedby` should be provided an id. We're saying if we have a message, the checkbox is described by that message, otherwise it should be undefined.

lukebennett88 (changes requested):

packages/checkbox/src/checkbox.tsx
freemany: Thanks @lukebennett88. Reverted.

freemany:
> I noticed that we have two Jest configs. Can you delete the old one as part of this PR please @freemany 🙏 https://github.com/brighte-labs/spark-web/blob/freeman/tests/packages/jest.config.js

I can move the inside one from packages to the root, but the displayName should be updated from 'spark-web-packages' to something just `spark-web-monorepo` or `spark-web`?

freemany:
>

freemany:
> I noticed that we have two Jest configs. Can you delete the old one as part of this PR please @freemany 🙏 https://github.com/brighte-labs/spark-web/blob/freeman/tests/packages/jest.config.js

We have multi packages setup in our repo, so each package has its own jest config file.

freemany:
> I noticed that we have two Jest configs. Can you delete the old one as part of this PR please @freemany 🙏 https://github.com/brighte-labs/spark-web/blob/freeman/tests/packages/jest.config.js

We have multi package setup in our repo, each package will have its own jest config file.

nderkim:
> > I noticed that we have two Jest configs. Can you delete the old one as part of this PR please @freemany 🙏 https://github.com/brighte-labs/spark-web/blob/freeman/tests/packages/jest.config.js
> 
> We have multi project setup in our repo, each project will have its own jest config file.

i think in this case we can just have 1 config at the root since every package should use the same config

freemany:
> > > I noticed that we have two Jest configs. Can you delete the old one as part of this PR please @freemany 🙏 https://github.com/brighte-labs/spark-web/blob/freeman/tests/packages/jest.config.js
> > 
> > 
> > We have multi project setup in our repo, each project will have its own jest config file.
> 
> i think in this case we can just have 1 config at the root since every package should use the same config

I think we can do that, updated.

freemany:
> I noticed that we have two Jest configs. Can you delete the old one as part of this PR please @freemany 🙏 https://github.com/brighte-labs/spark-web/blob/freeman/tests/packages/jest.config.js

Updated.

lukebennett88 (approved):

PR #18, opened 2022-04-18
Fix more linting and formatting errors

lukebennett88:


.eslintrc.js
```diff
@@ -4,7 +4,8 @@ const ERROR = 2;
 
 const reactComponentTypeMessage = {
   message:
-    'This type includes the children prop which is generally wrong, instead of using this type, type the props of the component',
+    'This type includes the children prop which is generally wrong, ' +
+    'instead of using this type, type the props of the component',
```
lukebennett88, line 8: Splitting up the comments in this file so they're under 80 characters and easier to read

packages/text-input/src/TextInput.tsx
```diff
@@ -60,13 +60,14 @@ export const TextInput = forwardRef<HTMLInputElement, TextInputProps>(
...
 );
+TextInput.displayName = 'TextInput';
 
 // Styled components
 // ------------------------------
 
 export type UseInputProps = Pick<FieldContextType, 'disabled' | 'invalid'>;
 
-export function useInput({ disabled, invalid }: UseInputProps) {
+export function useInput({ disabled }: UseInputProps) {
```
lukebennett88, line 70: I've removed the unused value here, but `invalid` is still required for `useInput`.
I did this for two reasons:
- Don't need to update every use of `useInput` to remove `invalid`
- We don't have any invalid styles yet, but we need to revisit them (e.g. if you focus on an input that is invalid, you get a green focus ring which looks wrong to me)

packages/text/src/defaultTextProps.tsx
```diff
@@ -16,12 +15,12 @@ const DefaultTextPropsContext = createContext<DefaultTextProps>({
...
 });
 
-export const DefaultTextPropsProvider: React.FC<DefaultTextProps> = ({
+export function DefaultTextPropsProvider({
   children,
   size,
   tone,
   weight,
-}) => {
+}: DefaultTextProps & { children: ReactNode }) {
```
nderkim, line 23: should we use `React.PropsWithChildren`?

lukebennett88: `PropsWithChildren` has children as optional which it's not in this case

.eslintrc.js
```diff
@@ -118,10 +128,27 @@ module.exports = {
...
       rules: {
-        // TypeScript already checks for the following things and they conflict with TypeScript
+        // TypeScript already checks for the following things and they cause conflicts
         'import/no-unresolved': OFF,
         'no-undef': OFF,
       },
     },
+    {
+      files: [
+        '**/__tests__/**/*',
```
freemany, line 138: we don't have __tests__ folder

lukebennett88: We don't right now, but might in the future.

packages/utils/src/useComposedRefs.ts
```diff
@@ -38,9 +38,12 @@ export function assignRef<RefValueType = any>(
...
-      assignRef(ref, node);
-    }
-  }, refs);
+  return useCallback(
+    (node: any) => {
+      for (const ref of refs) {
+        assignRef(ref, node);
+      }
+    },
+    [refs]
```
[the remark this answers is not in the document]
lukebennett88: Yeah I'm not 100% sure on this, starting to think it's not. Maybe we should override the ESLint rule here.

Or maybe we could use [this instead](https://github.com/Andarist/use-composed-ref/blob/main/src/index.ts)?

nderkim (approved):

PR #19, opened 2022-04-19
Add fieldset component

nderkim:
Screen Shot 2022-04-19 at 3 56 00 pm

### DIFF
--- a/packages/fieldset/src/fieldset.test.tsx
+++ b/packages/fieldset/src/fieldset.test.tsx
@@ -1,0 +1,21 @@
+import { cleanup, render } from '@testing-library/react';
+
+import { Fieldset } from './fieldset';
+
+describe('Fieldset component', () => {
+  afterEach(cleanup);
+
+  it('should have legend as first direct child of fieldset', () => {
+    // https://www.accessibility-developer-guide.com/examples/forms/grouping-with-fieldset-legend/#legend-must-be-a-direct-child-of-fieldset
+
+    const { container } = render(
+      <Fieldset legend="Test legend">Test content</Fieldset>
+    );
+
+    const fieldsetEl = container.firstElementChild!;
+    expect(fieldsetEl.tagName).toBe('FIELDSET');
+
+    const legendEl = fieldsetEl.firstElementChild!;
+    expect(legendEl.tagName).toBe('LEGEND');
+  });
+});
--- a/packages/fieldset/src/fieldset.tsx
+++ b/packages/fieldset/src/fieldset.tsx
@@ -1,0 +1,47 @@
+import { css } from '@emotion/css';
+import { Box } from '@spark-web/box';
+import { Text } from '@spark-web/text';
+import { useTheme } from '@spark-web/theme';
+import type { ReactNode, VFC } from 'react';
+import { forwardRef } from 'react';
+
+export type FieldsetProps = {
+  /** The form fields that comprise the set. */
+  children?: ReactNode;
+  /** An identifier which must be unique in the whole document. */
+  id?: string;
+  /**
+   * Provide a caption that describes the set of form fields.
+   * @see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/legend
+   */
+  legend?: string;
+  /** Sets the size of the gap between the legend and children. */
+  gap?: 'small' | 'medium' | 'large';
+};
+
+export const Fieldset = forwardRef<HTMLFieldSetElement, FieldsetProps>(
+  ({ children, id, legend, gap = 'small' }, ref) => (
+    <Box as="fieldset" ref={ref} id={id}>
+      {legend && (
+        <>
+          <Text as="legend" inline tone="neutral" weight="strong">
+            {legend}
+          </Text>
+          <Gap gap={gap} />
+        </>
+      )}
+      {children}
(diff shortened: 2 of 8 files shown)


packages/fieldset/src/fieldset.tsx
```diff
@@ -0,0 +1,35 @@
...
+   * @see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/legend
+   */
+  legend: string;
+};
+
+export const Fieldset = forwardRefWithAs<'fieldset', FieldsetProps>(
+  ({ as, children, id, legend, ...props }, ref) => {
+    return (
+      <Box as={as ?? 'fieldset'} ref={ref} id={id} {...props}>
+        <Stack gap="small">
```
lukebennett88, line 24: `<legend>` must be the first child of a `<fieldset>`
See more: https://www.accessibility-developer-guide.com/examples/forms/grouping-with-fieldset-legend/#legend-must-be-a-direct-child-of-fieldset

packages/fieldset/src/fieldset.tsx
```diff
@@ -0,0 +1,35 @@
...
+  /** An identifier which must be unique in the whole document. */
+  id?: BoxProps['id'];
+  /**
+   * Provide a caption that describes the set of form fields.
+   * @see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/legend
+   */
+  legend: string;
+};
+
+export const Fieldset = forwardRefWithAs<'fieldset', FieldsetProps>(
```
lukebennett88, line 20: I don't _think_ we need to be able to change the underlying element, I'm pretty sure it should _always_ be a `<fieldset>`.
`<legend>` should always be nested inside of a fieldset, so if we change the parent element it will likely cause accessibility issues.

packages/fieldset/src/fieldset.tsx
```diff
@@ -0,0 +1,35 @@
...
+  ({ as, children, id, legend, ...props }, ref) => {
```
lukebennett88, line 21: We're trying to avoid prop spreading.
See more: https://brighte.atlassian.net/wiki/spaces/SPARK/pages/2150793621/Open+Questions+Thinking

packages/fieldset/src/fieldset.tsx
```diff
@@ -0,0 +1,35 @@
...
+    return (
+      <Box as={as ?? 'fieldset'} ref={ref} id={id} {...props}>
```
lukebennett88, line 23: We're trying to avoid prop spreading.
See more: https://brighte.atlassian.net/wiki/spaces/SPARK/pages/2150793621/Open+Questions+Thinking

packages/fieldset/src/fieldset.test.tsx
```diff
@@ -0,0 +1,13 @@
+import { cleanup, render, screen } from '@testing-library/react';
+
+import { Fieldset } from './fieldset';
+
+describe('Fieldset component', () => {
+  afterEach(cleanup);
+
+  it('should show a fieldset with expected legend', () => {
```
lukebennett88, line 8: This is a pretty simple component, I'm not sure there is any value in checking if the component renders text into the legend.
Don't tell Freeman, but I think we should delete this test 🤫

packages/fieldset/src/fieldset.tsx
lukebennett88, line 24: The `gap` prop should get passed into this component so we can match the spacing of the rest of the form the fieldset is used in.

packages/fieldset/src/fieldset.tsx
```diff
@@ -0,0 +1,35 @@
...
+        <Stack gap="small">
+          <Box as="legend">
+            <Text inline tone={'neutral'} weight="strong">
```
lukebennett88, line 26: Nitpick: we don't need the curly quotes around `neutral`

packages/fieldset/src/fieldset.tsx
nderkim: nice 🎉

packages/fieldset/src/fieldset.tsx
nderkim: sorry, i thought this was the convention 🤦

packages/fieldset/src/fieldset.tsx
nderkim: sorry, i thought this was the convention 🤦

packages/fieldset/src/fieldset.test.tsx
nderkim: 🤭

packages/fieldset/src/fieldset.tsx
nderkim: agreed ✅

packages/fieldset/src/fieldset.tsx
lukebennett88: We actually are doing this in a bunch of places, but we want to roll it back

matildaPan (approved):
awesome work

lukebennett88 (approved):

lukebennett88 (approved):
I'm not entirely convinced that `legend` should be optional, but lets roll with it.

PR #31, opened 2022-04-20
Update tokens and add new colours page to docs

lukebennett88:
- First pass at updating tokens
- Remove temporary tokens
- Delete custom theme for docs (we don't need it anymore now that tokens are better)
- [Add new colours page to docs](https://spark-web-docs-git-luke-update-tokens-brighte.vercel.app/colours)


packages/theme/src/colors.ts
```diff
@@ -67,8 +65,7 @@ export const colors = {
     '100': '#fff5eb',
     '200': '#fdddc4',
     '300': '#face9b',
-    '400': '#f2b56d',
-    '500': '#d36815',
+    '500': '#ffaa44',
```
JedWatson, line 68: No `400`?

lukebennett88: `400` is reserved at the moment:
https://www.figma.com/file/0Guk4n4aEQIN0M0zHUSMDZ/Spark-Web-UI

JedWatson:
Looks good! love the new colours page in the docs.

🤔 it's a bit strange missing the `400` colour for a few of the scales? assuming that's a design thing?

Also, needs a changeset. This looks like a major

lukebennett88:
> 🤔 it's a bit strange missing the `400` colour for a few of the scales? assuming that's a design thing?

@jordangeizer is more across the reasoning than I am, but we don't need anything in this range, so I'm pretty sure we're just reserving those tokens for now, so that we _can_ add them later on if we end up needing something inbetween 300–500.

JedWatson (approved):

PR #40, opened 2022-04-24
Add @babel/plugin-transform-runtime

lukebennett88:
The SSR package was trying to use a global (`regeneratorRuntime`) which wasn't being defined:

next-error

[@babel/plugin-transform-runtime](https://babel.dev/docs/en/babel-plugin-transform-runtime#regenerator) adds this for us automatically (and it also claims to help reduce bundle size).

I've tested this locally and it seems to be working as expected.

I've also bumped the versions of our packages since Babel (which Preconstruct runs on all of our packages to create the builds) will output slightly different code now that we're using the transform runtime plugin.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -10,5 +10,5 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": []
+  "ignore": ["@spark-web/docs"]
 }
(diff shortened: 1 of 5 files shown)


.changeset/config.json
```diff
@@ -10,5 +10,5 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": []
+  "ignore": ["@spark-web/docs"]
```
lukebennett88, line 13: Snuck this into the PR since we don't want changesets to publish the docs

oscargws (approved):

JedWatson (approved):

PR #42, opened 2022-04-26
Fix multiple versions of React

lukebennett88:
When using Spark Web with a fresh install of Next.js (which uses React 18 by default) I get the following error:

PixelSnap 2022-04-26 at 16 03 50@2x

Looks like this is because we're shipping React 17.0.2 with the design system.

To address this by moving React _out_ of our `dependencies`, _into_ our `devDependencies` and adding `"react": ">=17.0.2"` to our `peerDependencies`.

I've also sorted the `package.json` files using [this npx script](https://www.npmjs.com/package/sort-package-json) to  make the find-and-replace a little easier.

### DIFF
--- a/.changeset/spicy-bees-reply.md
+++ b/.changeset/spicy-bees-reply.md
@@ -1,0 +1,38 @@
+---
+'@spark-web/a11y': patch
+'@spark-web/accordion': patch
+'@spark-web/alert': patch
+'@spark-web/analytics': patch
+'@spark-web/box': patch
+'@spark-web/button': patch
+'@spark-web/checkbox': patch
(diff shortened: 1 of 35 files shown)


JedWatson (changes requested):
@lukebennett88 can you break this PR up into two: one to clean up the package.json ordering (first) and then a separate one to just fix the dependencies?

hard to review at the moment

lukebennett88:
Sure thing @JedWatson

lukebennett88:
PR to sort packages here:
https://github.com/brighte-labs/spark-web/pull/43

jesstelford (approved):

PR #45, opened 2022-04-27
Float input component

justinfaynhan:
Added new float input component


lukebennett88:
Can you rebase from `main` @justinfaynhan 🙏
There are a few conflicts.

packages/float-input/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Uncontrolled
+
+A `FloatInput` can also be uncontrolled, managing it's own internal state. To access the value, instead of writing an onChange handler, you would use a ref to get form values from the DOM.
+
+```jsx live
+const inputRef = React.useRef(null);
+const [value, setValue] = React.useState('');
+const showValueHandler = React.useCallback(() => {
+  setValue(inputRef.current?.value)
```
justinfaynhan, line 62: Noting that the ref value is formatted i.e. 10000.101 -> 10,000.101

packages/float-input/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+A `FloatInput` can also be uncontrolled, managing it's own internal state. To
+access the value, instead of writing an onChange handler, you would use a ref to
+get form values from the DOM.
+
+```jsx live
+const inputRef = React.useRef(null);
+const [value, setValue] = React.useState('');
+const showValueHandler = React.useCallback(() => {
+  setValue(inputRef.current?.value);
```
justinfaynhan, line 73: Noting that the ref value is formatted i.e. 10000.101 -> 10,000.101 and always string. This differs from controlled flow where value is string | number.

docs/components/spark-components.ts
```diff
@@ -0,0 +1,30 @@
+export { Placeholder } from './example-helpers';
```
lukebennett88, line 1: We shouldn't need this file anymore. @nderkim and @jesstelford wrote a script that will generate the exports for us:
https://github.com/brighte-labs/spark-web/pull/34/files

justinfaynhan: Thanks for context, removed the file

packages/float-input/src/FloatInput.tsx
```diff
@@ -0,0 +1,37 @@
...
+      onFocus,
+    };
+
+    if (onChange !== undefined && value !== undefined) {
+      inputs = { ...inputs, onChange, value };
+    }
+
+    const floatInputProps = useFloatInput(inputs);
+
+    return <TextInput ref={ref} {...textInputProps} {...floatInputProps} />;
```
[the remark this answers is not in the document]
lukebennett88: We might be able to improve the experience for mobile users by adding the following attributes to our input:

```tsx
type="text"
pattern="[0-9]*"
inputMode="numeric"
```

Here are some articles suggesting it (as opposed to an `<input type="number" />`:
- https://bradfrost.com/blog/post/you-probably-dont-need-input-typenumber/
- https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/

Only thing I'm not sure about is if it is a good idea for floats (not just integers), maybe we should do some testing?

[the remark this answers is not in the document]
lukebennett88: As discussed, we probably want to expose some slots for adornments so we can use this for currency inputs and inputs with suffixes etc.
In this case we'll need to take the focus states off the text input and move them up to a wrapper element which we can style with the `:focus-within` pseudo selector (since an input cannot receive children).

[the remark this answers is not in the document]
justinfaynhan: Resolved prop spreading and the mentioned accessible input props. ✅

For adornments, I will resolve them in a later PR for the component inputs. I'd say this is fine to merge for now if no other issues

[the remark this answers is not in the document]
justinfaynhan: See https://github.com/brighte-labs/spark-web/pull/50 for :focus-within psedo selectors for adornments support

packages/text-input/src/TextInput.tsx
```diff
@@ -32,7 +40,8 @@ export type TextInputProps = {
    * How an input behaves varies considerably depending on the value of its type
    * attribute. If this attribute is not specified, the default type "text".
    */
-  type?: keyof typeof validTypes;
+  type?: validTypes;
+  mode?: validModes;
```
lukebennett88, line 44: Should be `inputmode`

packages/float-input/src/useFloatInput.ts
```diff
@@ -0,0 +1,163 @@
...
+
+type UseFloatInputUncontrolledProps = {
+  fractionDigits?: number;
+  onChange?: never;
+  value?: never;
+} & Pick<AllHTMLAttributes<HTMLInputElement>, 'onFocus' | 'onBlur'>;
+
+export type UseFloatInputProps =
+  | UseFloatInputControlledProps
+  | UseFloatInputUncontrolledProps;
```
lukebennett88, line 78: Theres a bit of double-up on the discriminated union here. Maybe something like this?

```suggestion
type ControlledFloatInputProps = {
  onChange: (value: number | string) => void;
  value: number | string;
};

type UncontrolledFloatInputProps = {
  onChange?: never;
  value?: never;
};

export type UseFloatInputProps = {
  fractionDigits?: number;
  onFocus?: AllHTMLAttributes<HTMLInputElement>['onFocus'];
  onBlur?: AllHTMLAttributes<HTMLInputElement>['onBlur'];
} & (ControlledFloatInputProps | UncontrolledFloatInputProps);
```

lukebennett88 (approved):
LGTM 🎉

PR #50, opened 2022-04-28
Feat/currency input component

justinfaynhan:
Implemented new currency input component. Includes changes to text input component to support starting and ending adornments.


packages/text-input/src/InputAdornment.tsx
```diff
@@ -0,0 +1,187 @@
...
+// Context
+// ------------------------------
+
+type InputAdornmentContextType = { placement: PlacementType };
+
+/**
+ * Components like the `SelectInput` may subscribe to the adornment context and
+ * change their appearance or behaviour.
+ */
+const InputAdornmentContext = createContext<InputAdornmentContextType | null>(
```
freemany, line 22: You may split this file into smaller ones as it is quite big and at least `Context` can be in its own file.

packages/text-input/src/TextInput.tsx
```diff
@@ -42,26 +45,97 @@ export type TextInputProps = {
...
+      baseline: false,
+      tone: disabled ? 'disabled' : 'neutral',
+      size: 'standard',
+      weight: 'regular',
+    });
 
-    return (
+    const [typographyStyles, responsiveStyles] = textStyles;
+    const { startAdornment, endAdornment } = childrenToAdornments(children);
+    const hasAdornments = Boolean(children);
```
lukebennett88, line 72: Rather than have the ternary here, I think we can get away with something like this:

```tsx
import { css } from '@emotion/css';
import { useFocusRing } from '@spark-web/a11y';
import { Box } from '@spark-web/box';
import type { FieldContextType } from '@spark-web/field';
import { useFieldContext } from '@spark-web/field';
import { useText } from '@spark-web/text';
import type { DataAttributeMap } from '@spark-web/utils/internal';
import { buildDataAttributes } from '@spark-web/utils/internal';
import type { AllHTMLAttributes } from 'react';
import { forwardRef } from 'react';

import type { AdornmentsAsChildren } from './childrenToAdornments';
import { childrenToAdornments } from './childrenToAdornments';

type validTypes =
  | 'text'
  | 'password'
  | 'email'
  | 'search'
  | 'number'
  | 'tel'
  | 'url';
type validModes =
  | 'none'
  | 'text'
  | 'tel'
  | 'url'
  | 'email'
  | 'numeric'
  | 'decimal'
  | 'search';

type NativeInputProps = Pick<
  AllHTMLAttributes<HTMLInputElement>,
  'onBlur' | 'onFocus' | 'onChange' | 'placeholder' | 'value'
>;

export type TextInputProps = {
  /** Map of data attributes. */
  data?: DataAttributeMap;
  /**
   * How an input behaves varies considerably depending on the value of its type
   * attribute. If this attribute is not specified, the default type "text".
   */
  type?: validTypes;
  mode?: validModes;
  /**
   * Adorn the input with ornamental element(s) to aid user input, or
   * interactive element(s) to augment user input. Each child **must** be
   * wrapped with the `InputAdornment` component to ensure proper layout,
   * otherwise it will not be rendered.
   */
  children?: AdornmentsAsChildren;
} & NativeInputProps;

/** Organize and emphasize information quickly and effectively in a list of text elements. */
export const TextInput = forwardRef<HTMLInputElement, TextInputProps>(
  ({ children, data, ...consumerProps }, forwardedRef) => {
    const { disabled, invalid, ...a11yProps } = useFieldContext();
    const theme = useTheme();
    const focusRingStyles = useFocusRing({ always: true });
    const textStyles = useText({
      baseline: false,
      tone: disabled ? 'disabled' : 'neutral',
      size: 'standard',
      weight: 'regular',
    });

    const [typographyStyles, responsiveStyles] = textStyles;
    const { startAdornment, endAdornment } = childrenToAdornments(children);

    return (
      <Box
        background={disabled ? 'inputDisabled' : 'input'}
        border={invalid ? 'critical' : 'field'}
        borderRadius="small"
        display="inline-flex"
        alignItems="center"
        flexDirection="row"
        height="medium"
        marginY="none"
        className={css({
          ':focus-within': {
            ...focusRingStyles,
            borderColor: theme.border.color.fieldAccent,
          },
        })}
      >
        {startAdornment}
        <Box
          as="input"
          ref={forwardedRef}
          disabled={disabled}
          // Styles
          flex={1}
          height="medium"
          paddingX="medium"
          paddingLeft={startAdornment ? 'none' : 'medium'}
          paddingRight={endAdornment ? 'none' : 'medium'}
          className={css({
            ...typographyStyles,
            ...responsiveStyles,
            ':focus': {
              outline: 'none',
            },
            ':enabled': {
              '&:hover': {
                borderColor: theme.border.color.fieldHover,
              },
            },
          })}
          {...(data ? buildDataAttributes(data) : null)}
          {...a11yProps}
          {...consumerProps}
        />
        {endAdornment}
      </Box>
    );
  }
);
TextInput.displayName = 'TextInput';

// Styled components
// ------------------------------

export type UseInputProps = Pick<FieldContextType, 'disabled' | 'invalid'>;

export function useInput({ disabled }: UseInputProps) {
  const theme = useTheme();
  const focusRingStyles = useFocusRing({ always: true });
  const textStyles = useText({
    baseline: false,
    tone: disabled ? 'disabled' : 'neutral',
    size: 'standard',
    weight: 'regular',
  });

  const [typographyStyles, responsiveStyles] = textStyles;

  return {
    ...typographyStyles,
    ...responsiveStyles,
    ':enabled': {
      '&:hover': {
        borderColor: theme.border.color.fieldHover,
      },
      '&:focus': {
        ...focusRingStyles,
        borderColor: theme.border.color.fieldAccent,
      },
    },
  };
}

```

Ideally the styles should also be abstracted to a hook. I know we're sharing this for the Select and TextArea etc, so we might need to provider a prop to differentiate the use-cases.

.changeset/chilled-foxes-compete.md
```diff
@@ -0,0 +1,11 @@
+---
+'@spark-web/currency-input': major
+'@spark-web/docs': minor
+'@spark-web/field': minor
+'@spark-web/float-input': minor
+'@spark-web/text-input': minor
+---
+
+Added new currency input component!! Some changes have been made to its
```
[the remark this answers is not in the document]
justinfaynhan: Oops, will fix !!!!!!

packages/currency-input/src/currencySymbolMap.ts
```diff
@@ -0,0 +1,192 @@
+export type CurrencyType = keyof typeof currencySymbolMap;
+
+export const currencySymbolMap = {
```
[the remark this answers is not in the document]
justinfaynhan: I added extra currency support mindful that we do have future plans to expand geographies (i.e. europe). Also think this is a low-hanging enough of a fruit to add into this component.

packages/text-input/src/TextInput.tsx
justinfaynhan: I've simplified the component and removed the ternary

packages/text-input/src/childrenToAdornments.tsx
```diff
@@ -0,0 +1,48 @@
...
+        endAdornment = child;
+      }
+      if (child.props.placement === 'start') {
+        startAdornment = child;
+      }
+    }
+  });
+
+  return { endAdornment, startAdornment };
+};
```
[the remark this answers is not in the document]
justinfaynhan: Have incorporated your suggestions into the file and removed the AdornmentPlaceholder ✅

.changeset/chilled-foxes-compete.md
```diff
@@ -0,0 +1,11 @@
+---
+'@spark-web/currency-input': major
+'@spark-web/docs': minor
```
lukebennett88, line 3: Nitpick: I wouldn't worry about bumping the docs version

packages/currency-input/src/index.tsx
```diff
@@ -0,0 +1,2 @@
+export type { CurrencyInputProps } from './CurrencyInput';
```
lukebennett88, line 1: Nitpick: file extension can be `.ts` since there is no JSX here, and the other `src/index.ts` files seperate the types from the code with a `// types` comment.

packages/text-input/src/TextInput.tsx
```diff
@@ -41,42 +44,89 @@ export type TextInputProps = {
...
+        flexDirection="row"
         height="medium"
-        paddingX="medium"
-        className={css(inputStyles)}
-        {...(data ? buildDataAttributes(data) : null)}
-        {...a11yProps}
-        {...consumerProps}
-      />
+        marginY="none"
+        className={css({
```
lukebennett88, line 86: We should abstract the style for the input and the wrapper into a hook like we do elsewhere to keep it consistent.

packages/text-input/src/TextInput.tsx
```diff
@@ -41,42 +44,89 @@ export type TextInputProps = {
...
+          flex={1}
+          height="medium"
+          paddingX="medium"
+          paddingLeft={startAdornment ? 'none' : 'medium'}
+          paddingRight={endAdornment ? 'none' : 'medium'}
+          className={css({
+            ...typographyStyles,
+            ...responsiveStyles,
+            ':focus': {
+              outline: 'none',
```
lukebennett88, line 108: Instead of outline: 'none' we should change to:
```tsx
outline: "2px solid transparent",
outlineOffset: "2px"
```

This ensures elements are still visibly focused for Windows high contrast mode:
<https://tailwindcss.com/docs/outline-style#removing-outlines>

packages/text-input/src/TextInput.tsx
justinfaynhan: How should we go about the existing useInput hook used by the other components (select, textarea, etc) now that the hook isn't actively used by the textinput component. I don't think it will be as straightfowrard just replacing the old hook with the new style hooks which are broken up in wrapper and child parts. Should we just leave it for compatibility with the other components?

justinfaynhan: I've added the new style classname hooks for the input wrapper and nested component alongside the existing old input style hook

packages/text-input/src/InputAdornment.tsx
freemany: One reason besides others is to make the same responsibility into one file, that file can be targeted for a test, like if you have `foo.ts` and `bar.ts`, it intends or reminds us to test it with `foo.test.ts` and `bar.test.ts`. But my comment is not a blocker.

packages/text-input/src/InputAdornment.tsx
```diff
@@ -0,0 +1,124 @@
...
+  end: {
+    paddingLeft: 'xsmall',
+    paddingRight: 'medium',
+  },
+} as const;
+
+type PlacementType = keyof typeof placementToPadding;
+export type InputAdornmentProps = {
+  children: ReactElement;
+  /**
```
matildaPan, line 35: should we move all these comments to readme as part of docs?

justinfaynhan: Good point, InputAdornment is part of the textinput API but it's not currently documented

justinfaynhan: There is a section already in TextInput readme providing an example of how to use the InputAdornment field with the TextInput component. I think that should be sufficient explicit documentation.

lukebennett88:
Nitpick: Can you run `npx sort-package-json` in the `packages/float-input` directory so it's in the same order as everything else 🙏

packages/currency-input/src/CurrencyInput.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { currencySymbolMap } from './currencySymbolMap';
+
+/** NOTE: The "start" adornment is filled, so we can only accept a single child element. */
+export type CurrencyInputProps = {
+  currencyType?: CurrencyType;
+} & Omit<FloatInputProps, 'fractionDigits'>;
+
+/** A component for inputting numbers into the app via a keyboard. Enforces 2 fraction digits. */
+export const CurrencyInput = forwardRef<HTMLInputElement, CurrencyInputProps>(
+  ({ currencyType, ...props }, ref) => {
```
lukebennett88, line 21: Nitpick: can we rename `ref` to `forwardedRef` to be a bit more specific about what sort of ref it is 🙏

packages/float-input/src/FloatInput.tsx
```diff
@@ -6,11 +6,20 @@ import type { UseFloatInputProps } from './useFloatInput';
 import { useFloatInput } from './useFloatInput';
 
 export type FloatInputProps = UseFloatInputProps &
-  Omit<TextInputProps, 'onChange' | 'value' | 'type' | 'mode' | 'children'>;
+  Omit<TextInputProps, 'onChange' | 'value' | 'type' | 'mode'>;
```
lukebennett88, line 9: Should be `inputMode` not `mode`

packages/text-input/src/TextInput.tsx
```diff
@@ -101,4 +122,45 @@ export function useInput({ disabled }: UseInputProps) {
...
+
+  return {
+    ':focus-within': {
+      ...focusRingStyles,
+      borderColor: theme.border.color.fieldAccent,
+    },
+  };
+};
+
+export const useInputNested = ({ disabled }: UseInputProps) => {
```
lukebennett88, line 139: Feels like there is a lot of duplication in this hook.
We can tuck this all away inside of the `useInput` hook and just add an `isNested` option to the config object that defaults to false so it doesn't break the other components that use it.

Something like this should work:

```tsx
export type UseInputProps = Pick<FieldContextType, 'disabled' | 'invalid'> & {
  isNested?: boolean;
};

export const useInput = ({ disabled, isNested = false }: UseInputProps) => {
  const theme = useTheme();
  const focusRingStyles = useFocusRing({ always: true });
  const textStyles = useText({
    baseline: false,
    tone: disabled ? 'disabled' : 'neutral',
    size: 'standard',
    weight: 'regular',
  });

  const [typographyStyles, responsiveStyles] = textStyles;

  return {
    ...typographyStyles,
    ...responsiveStyles,
    ...(isNested
      ? {
          ':focus': {
            // This removes the nested input outline visibility since
            // the wrapper will be outlined, but still visibly focusable
            // to windows high contrast mode users.
            // @see https://tailwindcss.com/docs/outline-style#removing-outlines
            outline: '2px solid transparent',
            outlineOffset: '2px',
          },
        }
      : {
          ':enabled': {
            '&:hover': {
              borderColor: theme.border.color.fieldHover,
            },
            '&:focus': {
              ...focusRingStyles,
              borderColor: theme.border.color.fieldAccent,
            },
          },
        }),
  } as const;
};
```

packages/text-input/src/TextInput.tsx
```diff
@@ -101,4 +122,45 @@ export function useInput({ disabled }: UseInputProps) {
...
+      // to windows high contrast mode users.
+      // @see https://tailwindcss.com/docs/outline-style#removing-outlines
+      outline: '2px solid transparent',
+      outlineOffset: '2px',
+    },
+    ':enabled': {
+      '&:hover': {
+        borderColor: theme.border.color.fieldHover,
+      },
+    },
```
lukebennett88, line 164: I don't think this does anything, I'm pretty sure we can remove it.

justinfaynhan:
> Nitpick: Can you run `npx sort-package-json` in the `packages/float-input` directory so it's in the same order as everything else 🙏

@lukebennett88 I suggest we add in an [appropriate prettier plugin](https://github.com/matzkoh/prettier-plugin-packagejson) for this rule and have it run as a pre-commit hook for better dev experience. I can make a PR later

packages/text-input/src/TextInput.tsx
justinfaynhan: Suggestion added ✅

packages/currency-input/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "license": "MIT",
+  "main": "dist/spark-web-currency-input.cjs.js",
+  "module": "dist/spark-web-currency-input.esm.js",
+  "files": [
+    "dist"
+  ],
+  "dependencies": {
+    "@babel/runtime": "^7.14.6",
+    "@emotion/css": "^11.7.1",
+    "@spark-web/box": "^1.0.0",
```
lukebennett88, line 13: Nitpick: Looks like we're not using the `@spark-web/box` so we can remove it.
Maybe consider bumping the other packages to the most recent version as well?
This should work:
```
"dependencies": {
  "@babel/runtime": "^7.14.6",
  "@emotion/css": "^11.7.1",
  "@spark-web/field": "^1.0.3",
  "@spark-web/float-input": "^0.0.0",
  "@spark-web/text": "^1.0.3",
  "@spark-web/text-input": "^1.0.3"
},
```

packages/text-input/src/TextInput.tsx
```diff
@@ -42,41 +45,61 @@ export type TextInputProps = {
...
-// Styled components
-// ------------------------------
-
-export type UseInputProps = Pick<FieldContextType, 'disabled' | 'invalid'>;
+export type UseInputProps = Pick<FieldContextType, 'disabled' | 'invalid'> & {
+  isNested?: boolean;
+};
 
-export function useInput({ disabled }: UseInputProps) {
+export const useInput = ({ disabled, isNested = false }: UseInputProps) => {
```
lukebennett88, line 102: Looks like we've lost our focus rings. Need to add `:focus-within` back:
https://spark-web-docs-gemv80e9t-brighte.vercel.app/package/text-input

lukebennett88 (approved):
Nice one, looks good to me! 🚢

PR #51, opened 2022-04-29
Test Textarea

freemany:
Screen Shot 2022-04-29 at 2 02 19 pm

### DIFF
--- a/packages/text-area/src/text-area.test.tsx
+++ b/packages/text-area/src/text-area.test.tsx
@@ -1,22 +1,18 @@
 import '@testing-library/jest-dom';
 
-import type { Tone } from '@spark-web/field/src';
-import { Field } from '@spark-web/field/src';
+import type { FieldProps } from '@spark-web/field';
+import { Field } from '@spark-web/field';
 import { render, screen } from '@testing-library/react';
 
+import type { TextAreaProps } from '.';
 import { TextArea } from './text-area';
 
 const renderComponent = (
-  feildProps: {
-    label: string;
-    disabled?: boolean;
-    message?: string;
-    tone?: Tone;
-  },
-  props?: { data: { [key: string]: string } }
+  fieldProps: Omit<FieldProps, 'children'>,
+  props?: TextAreaProps
 ) => {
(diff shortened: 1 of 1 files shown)


packages/text-area/src/text-area.test.tsx
```diff
@@ -0,0 +1,70 @@
+import '@testing-library/jest-dom';
+
+import type { Tone } from '@spark-web/field/src';
+import { Field } from '@spark-web/field/src';
+import { render, screen } from '@testing-library/react';
+
+import { TextArea } from './text-area';
+
+const renderComponent = (
+  feildProps: {
```
lukebennett88, line 10: typo: should be `fieldProps`

packages/text-area/src/text-area.test.tsx
```diff
@@ -0,0 +1,70 @@
+import '@testing-library/jest-dom';
+
+import type { Tone } from '@spark-web/field/src';
+import { Field } from '@spark-web/field/src';
```
lukebennett88, line 4: ☝️ We shouldn't need the `/src` at the end of these imports

packages/text-area/src/text-area.test.tsx
lukebennett88: suggestion: `fieldProps: Omit<FieldProps, 'children'>` instead of declaring the types ourself?

packages/text-area/src/text-area.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { TextArea } from './text-area';
+
+const renderComponent = (
+  feildProps: {
+    label: string;
+    disabled?: boolean;
+    message?: string;
+    tone?: Tone;
+  },
+  props?: { data: { [key: string]: string } }
```
lukebennett88, line 16:
```suggestion
textAreaProps?: TextAreaProps
```

lukebennett88 (commented):
Looks good, just a few nitpick suggestions

lukebennett88 (approved):

PR #52, opened 2022-05-01
Add files array to package.json

lukebennett88:
Currently we're shipping the entire content of the package directory with all of our packages:
<https://unpkg.com/browse/@spark-web/button@1.0.3/>
We only need to package the contents of the `dist` directory, so adding it to every `package.json` file in the `packages` directory should dramatically reduce the library size.

See: <https://docs.npmjs.com/cli/v8/configuring-npm/package-json#files>

### DIFF
--- a/packages/float-input/package.json
+++ b/packages/float-input/package.json
@@ -4,6 +4,9 @@
   "license": "MIT",
   "main": "dist/spark-web-float-input.cjs.js",
   "module": "dist/spark-web-float-input.esm.js",
+  "files": [
+    "dist"
+  ],
   "dependencies": {
     "@babel/runtime": "^7.14.6",
     "@emotion/css": "^11.7.1",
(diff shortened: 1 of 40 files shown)


packages/float-input/package.json
```diff
@@ -1,17 +1,9 @@
 {
   "name": "@spark-web/float-input",
-  "license": "MIT",
   "version": "0.0.0",
+  "license": "MIT",
   "main": "dist/spark-web-float-input.cjs.js",
   "module": "dist/spark-web-float-input.esm.js",
```
nderkim, line 6: did you get distracted doing the sort and forget to do the files thingamajiggy here?

nderkim (approved):

packages/float-input/package.json
lukebennett88: Yeah I must have missed this after rebasing (this is a new package).
Good catch 👍

PR #54, opened 2022-05-02
style: added pre-commit linting hook

justinfaynhan:
It is common to forget to lint/format code before pushing changes and breaking PR build checks, resulting in roundabout lint fix commits. Adding precommit hook linting for better developer experience.


matildaPan (approved):
awesome

PR #55, opened 2022-05-02
Add VSCode workspace settings

lukebennett88:
Add VSCode workspace extensions and recommended settings.
Mostly the same as what we had in the energy monorepo, lightly edited with some settings that I've found to be more reliable ([borrowed from here](https://github.com/wesbos/eslint-config-wesbos#with-vs-code)).

### DIFF
--- a/.vscode/settings.json.example
+++ b/.vscode/settings.json.example
@@ -1,0 +1,19 @@
+{
+  "editor.codeActionsOnSave": {
+    "source.fixAll": true
+  },
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "editor.formatOnSave": true,
+  "eslint.packageManager": "yarn",
+  "eslint.validate": [
+    "javascript"
+  ],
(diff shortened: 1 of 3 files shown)


.vscode/settings.json.example
```diff
@@ -0,0 +1,19 @@
+{
```
lukebennett88, line 1: Not sure if this should end in `.example` or not (this is how it was in the energy monorepo).
I don't think these settings are applied as they are, so it's more of a list of _suggestions_. If it was `settings.json` it would probably override you global settings which some of us might not want if we have our editors configured _just so_.

freemany: This is an example, can it be removed?

freemany (approved):

matildaPan (approved):

.vscode/settings.json.example
lukebennett88: It's a good reference for format-on-save if it isn't working for you.

PR #58, opened 2022-05-02
docs: updated all bitbucket reference links in component README docs

justinfaynhan:
Package README.md links were still referring to the older bitbucket repo links prior to github opensourcing. This PR fixes those link and some minor fixes to broken links.


lukebennett88:
@justinfaynhan I can't remember — what is the reasoning for linking to a specific commit instead of a the most recent version in `main`?

packages/alert/README.md
```diff
@@ -87,4 +87,4 @@ In rare cases, you may need to provide a custom icon. To do so, we expose an
 | heading     | string                                          |               | Sets a heading for the alert.                                                                                                                                                    |
 
 [data-attribute-map]:
-  https://bitbucket.org/brighte-energy/energy/src/14a694872cc43bb454981bada65f5f12b56f77c9/spark-web/packages/utils-spark/src/buildDataAttributes.ts#spark-web/packages/utils-spark/src/buildDataAttributes.ts-1
+  https://github.com/brighte-labs/spark-web/blob/e7f6f4285b4cfd876312cc89fbdd094039aa239a/packages/utils/src/internal/buildDataAttributes.ts#L1
```
nderkim, line 90: Is this auto generated? If not, what is `e7f6f4285b4cfd876312cc89fbdd094039aa239a` pointing to?

justinfaynhan:
@nderkim The github links are pointing to the commit SHA that added the specific changes.

@lukebennett88 My rationale of choosing the commit SHA instead of main is that there may often be drift between what is on main vs what has been published and being used by consumers. If we have incoming breaking changes on main, then those links will potentially be broken in the docs. Instead, we pin it to the exact commit SHA that introduced those changes and avoid the drift in docs and source files.

nderkim:
ah because we have line number as part of the link as well 👍

packages/columns/README.md
```diff
@@ -89,8 +89,8 @@ listed here (excluding `display`, `alignItems`, `gap`, `flexDirection`,
 Extra props are passed into the underlying [`Box`](/package/box) component.
```
lukebennett88, line 89: Technically not part of this PR, but would you mind fixing up the typo on line `81`?
`[ResponsiveRangeProps][responsive-range-props]('below')` should be `[ResponsiveRangeProps][responsive-range-props]['below']`

packages/float-input/README.md
```diff
@@ -116,5 +116,5 @@ return (
 | onFocus?        | FocusEventHandler\<HTMLInputElement\> |         | Handler when input element is focused on.                                                                                                                   |
 | onBlur?         | FocusEventHandler\<HTMLInputElement\> |         | Handler when input element is blurred.                                                                                                                      |
 
-Additional props also include [`TextField`](/package/text-field) props which are
+Additional props also include [`TextLink`](/package/text-link) props which are
```
lukebennett88, line 119: Shouldn't this be `TextInput` instead of `TextLink`?

justinfaynhan: Good catch, I was fixing an existing typo but made another typo :)

lukebennett88 (approved):

PR #62, opened 2022-05-03
Flexsearch with search index and search component

matildaPan:
Ticket: https://brighte.atlassian.net/browse/SPRK-33
This PR is about: 
- export search index using flexsearch 
- search component using flexsearch API 

Remaining work will be covered by https://brighte.atlassian.net/browse/SPRK-62
Screen Shot 2022-05-03 at 4 31 09 pm


docs/package.json
```diff
@@ -4,11 +4,11 @@
   "private": true,
   "license": "MIT",
   "scripts": {
-    "build": "yarn playroom:build && INIT_CWD=$PWD next build  && ./scripts/generate-search-index.js && yarn storybook:build ",
+    "build": "yarn playroom:build && INIT_CWD=$PWD next build && ./scripts/generate-search-index.js && yarn storybook:build ",
     "contentlayer:build": "INIT_CWD=$PWD yarn contentlayer build",
     "dev": "yarn run search-generation && INIT_CWD=$PWD next dev",
     "lint": "next lint",
-    "playroom:build": "playroom build",
+    "playroom:build": "./scripts/generate-spark-components.js && playroom build",
```
matildaPan, line 11: `./scripts/generate-spark-components.js ` will be executed twice in `build`.

docs/components/header.tsx
```diff
@@ -4,15 +4,21 @@ import { Box } from '@spark-web/box';
 import { Container } from '@spark-web/container';
 import { Hidden } from '@spark-web/hidden';
 import { MenuIcon, XIcon } from '@spark-web/icon';
+import { Inline } from '@spark-web/inline';
 import { Link } from '@spark-web/link';
 import { Strong, Text } from '@spark-web/text';
 import { useTheme } from '@spark-web/theme';
+// @ts-expect-error
```
matildaPan, line 11: flexsearch has a bug

docs/components/header.tsx
```diff
@@ -143,3 +153,49 @@ const GitHubLink = () => {
...
+        await flexsearchDoc.import(element.key, JSON.parse(element.data));
+      });
+
+      flexsearchRef.current = flexsearchDoc;
+    };
+
+    fetchSearchIndex();
+  });
+
+  // Removing search for now (see https://brighte.atlassian.net/browse/SPRK-62)
```
matildaPan, line 179: This PR is only to implement search, we will integrate search with auto complete combo box after it is available. More info https://brighte.atlassian.net/browse/SPRK-62

docs/pages/_app.tsx
```diff
@@ -1,6 +1,9 @@
 import { SparkProvider } from '@spark-web/core';
 import { UniversalNextLink } from '@spark-web/next-utils';
+import type { SearchIndexType } from 'components/header';
 import { allPackages } from 'contentlayer/generated';
+// @ts-expect-error
```
matildaPan, line 5: flexsearch has a bug

docs/pages/_app.tsx
```diff
@@ -41,6 +46,32 @@ App.getInitialProps = async () => {
         .filter(({ slug }) => !ignorePackages.includes(slug))
         .map(({ title, slug }) => ({ name: title, href: `/package/${slug}` })),
     ],
+    searchIndex: (() => {
```
matildaPan, line 49: reason for having search index built in server side is because it can reduce the size of data source , and when search triggered , search index is sufficient .

docs/package.json
```diff
@@ -75,6 +77,7 @@
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
     "react-simple-code-editor": "^0.11.0",
+    "react-use-flexsearch": "^0.1.1",
```
nderkim, line 80: i think you can delete this?

docs/package.json
```diff
@@ -85,6 +88,7 @@
     "@storybook/addon-links": "^6.4.21",
     "@storybook/addon-postcss": "^2.0.0",
     "@storybook/react": "^6.4.21",
+    "@types/flexsearch": "^0.7.3",
```
nderkim, line 91: i think you can delete this too. Maybe this was causing the typescript bug?

docs/pages/_app.tsx
```diff
@@ -41,6 +46,32 @@ App.getInitialProps = async () => {
...
+      // @ts-expect-error
+      const flexSearchDocument = new FlexSearchDocument({
+        document: {
+          id: 'slug',
+          index: ['content'],
+        },
+      });
+
+      allPackages.forEach(pkg => {
```
jesstelford, line 58: `getInitialProps()` runs on both the server and the client. Meaning the client is now downloading the entire `allPackages` array to be able to generate the search index for every client side route change.

jesstelford: Looks like I accidentally introduced this as a bug for the navigation in https://github.com/brighte-labs/spark-web/commit/a98f5107eb2582c6f1808600b9b1435f8cbf1cff#diff-40d08f613c794b0baed1dd9274a7ac7ecf54dcfb8ae998ac612f70d1cb8c7a43R40

docs/pages/_app.tsx
```diff
@@ -41,6 +46,32 @@ App.getInitialProps = async () => {
...
+
+      const searchIndex: SearchIndexType = [];
+
+      flexSearchDocument.export(function (key: string, data: string) {
+        // you need to store both the key and the data!
+        // e.g. use the key for the filename and save your data
+        searchIndex.push({ key, data });
+      });
+
+      return searchIndex;
```
jesstelford, line 73: By returning the searchIndex here, it means every page load must pay the cost of loading the entire index.

The original goal was to lazy-load that search index only once the user clicks into the search box, thereby incurring the cost only if they do want to search (the assumption being that only a small fraction of page loads actually want to perform a search).

docs/pages/_app.tsx
nderkim: @jesstelford didn't we conclude that it only disables automatic static builds, but you can force it back by having an explicit `getStaticProps`?

jesstelford (approved):

PR #65, opened 2022-05-04
Update typography

lukebennett88:


packages/core/src/fonts.tsx
```diff
@@ -1,32 +1,12 @@
...
+ * TODO: the CSS file that loads Aestetico is a bit of a hack as it relies on the atomic deploys never changing
  * we should host this file on the brighte.com.au domain (or subdomain) in the future
  */
-export function AestheticoStylesheet(): JSX.Element {
+export function AesteticoStylesheet(): JSX.Element {
   return (
     <link
       rel="stylesheet"
-      href="https://spark-web-docs-5at71kbus-brighte.vercel.app/_next/static/css/b1c6707967e86a27.css"
+      href="https://spark-web-docs-a22gianxx-brighte.vercel.app/fonts.css"
```
lukebennett88, line 9: I've updated the stylesheet to only have two weights and spell the font correctly 🤦‍♂️
I've been spelling "Aesthetico" with a "H" this whole time 🙃

packages/core/src/fonts.tsx
```diff
@@ -1,32 +1,12 @@
...
     />
   );
 }
-
-export function InterStylesheet(): JSX.Element {
```
lukebennett88, line 14: Removing Inter to better align with the rest of Brighte's branding

packages/theme/package.json
```diff
@@ -10,7 +10,6 @@
   "dependencies": {
     "@babel/runtime": "^7.14.6",
     "@capsizecss/core": "^3.0.0",
-    "@capsizecss/metrics": "^0.1.0",
```
lukebennett88, line 13: We don't need this anymore since we were only using it for Inter

packages/theme/src/tokens.ts
```diff
@@ -35,23 +34,22 @@ export const defaultTokens = {
   typography: {
     fontFamily: {
       sans: {
-        fontMetrics: interFontMetrics as FontMetrics,
-        name: '"Inter", system-ui, -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", sans-serif',
+        fontMetrics: AesteticoFontMetrics,
+        name: '"Aestetico", ui-sans-serif, system-ui, -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, "Noto Sans", sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji"',
```
lukebennett88, line 38: Updated the fallbacks to match Tailwind

packages/theme/src/tokens.ts
```diff
@@ -35,23 +34,22 @@ export const defaultTokens = {
...
-        name: '"Aesthetico", "Inter", system-ui, -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", sans-serif',
+        fontMetrics: AesteticoFontMetrics,
+        name: '"Aestetico", ui-sans-serif, system-ui, -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, "Noto Sans", sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji"',
       },
     },
     fontWeight: {
       regular: 400,
-      medium: 500,
-      strong: 700,
+      semibold: 600,
```
lukebennett88, line 47: We've only got two weights now: `regular` and `semibold`.

packages/theme/src/tokens.ts
```diff
@@ -35,62 +34,61 @@ export const defaultTokens = {
...
         weak: 'regular',
-        regular: 'medium',
+        regular: 'semibold',
       },
       level: {
         '1': {
           mobile: {
-            fontSize: 28,
+            fontSize: 35,
             rows: 9,
```
lukebennett88, line 58: @jordangeizer I think we agreed on having responsive sizes for `h1` but not the rest?
If so, I couldn't find in Figma what sizes to use here.

jordangeizer: Just reviewed this @lukebennett88! Realised a `35` h1 is perfectly fine for mobile. So will need responsive type styles for h1 when for the _marketing_ theme.

lukebennett88: 👌

ChristopherMitchell242 (commented):
Should we add tone & weight to the Text component README props table. No examples of the allowed values there at the moment.

lukebennett88:
> Should we add tone & weight to the Text component README props table. No examples of the allowed values there at the moment.

Good point, will update 👍

ChristopherMitchell242 (approved):

PR #66, opened 2022-05-04
Fix/sprk 58 fix math error in dropzone component

ChristopherMitchell242:
This PR consists of two changes. 

The first change updates the number of bytes in a kB from 1024 to 1000. 
The second change updates the calculation to provide the max and min file size to the Dropzone component in Bytes (https://react-dropzone.js.org/). 

A subsequent ticket will be raised to cover the testing of this component as agreed with Luke.

### DIFF
--- a/packages/dropzone/src/dropzone.tsx
+++ b/packages/dropzone/src/dropzone.tsx
@@ -372,7 +372,7 @@ function useDropzoneStyles({
 
 function formatFileSize(numKb: number): string {
   if (numKb < 1000) {
-    return `${Math.round(numKb).toFixed(1)}kB`;
+    return `${Math.round(numKb).toFixed()}kB`;
   }
-  return `${Math.round(numKb / 1000).toFixed(1)}MB`;
+  return `${Math.round(numKb / 1000).toFixed()}MB`;
 }
--- a/packages/dropzone/src/dropzone.tsx
+++ b/packages/dropzone/src/dropzone.tsx
@@ -40,9 +40,9 @@ export type DropzoneProps = InputProps & {
   accept?: AcceptedType | AcceptedType[];
   /** Maximum number of files that the Dropzone should be allowed to accept. */
   maxFiles?: number;
-  /** Maximum file size that the Dropzone should be allowed to accept. */
+  /** Maximum file size that the Dropzone should be allowed to accept. Value should be provided in kB */
   maxFileSizeKb?: number;
-  /** Minimum file size that the Dropzone should be allowed to accept. */
+  /** Minimum file size that the Dropzone should be allowed to accept. Value should be provided in kB */
   minFileSizeKb?: number;
   /** When true, renders an image preview next to file previews. */
   showImageThumbnails?: boolean;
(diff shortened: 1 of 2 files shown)


lukebennett88:
Thanks @ChristopherMitchell242!
We're using [changesets](https://github.com/changesets/changesets) for managing our versioning/changelogs.
Do you mind adding a changeset to this PR? 🙏
You can either run `yarn changeset` and follow the prompts in the CLI, or use the link in the changeset-bot above to add one.

packages/dropzone/src/dropzone.tsx
```diff
@@ -96,8 +96,8 @@ export const Dropzone = React.forwardRef<HTMLInputElement, DropzoneProps>(
     } = useDropzone({
       accept,
       maxFiles,
-      maxSize: maxFileSizeKb && maxFileSizeKb / 1024,
-      minSize: minFileSizeKb && minFileSizeKb / 1024,
+      maxSize: maxFileSizeKb && maxFileSizeKb * 1000,
+      minSize: minFileSizeKb && minFileSizeKb * 1000,
```
nderkim, line 100: good catch 👍

nderkim: can i ask what the reason is for changing this to 1000?

lukebennett88: @nderkim here is the ticket: https://brighte.atlassian.net/browse/SPRK-58?atlOrigin=eyJpIjoiMzhkMjQ2NTNkNzc2NGNjYTgxNDRkMTRiZDg2ZTg5Y2EiLCJwIjoiaiJ9

Turns out we were using [kiB](https://simple.wikipedia.org/wiki/Kibibyte) not [kB](https://simple.wikipedia.org/wiki/Kilobyte)

nderkim:
you can probably just run `yarn format:write` to fix that ☝️

packages/dropzone/src/dropzone.tsx
nderkim: nevermind, @lukebennett88 explained it to me.. carry on 😅

nderkim (approved):

packages/dropzone/src/dropzone.tsx
```diff
@@ -371,8 +371,8 @@ function useDropzoneStyles({
 }
 
 function formatFileSize(numKb: number): string {
-  if (numKb < 1024) {
+  if (numKb < 1000) {
     return `${Math.round(numKb).toFixed(1)}kB`;
```
freemany, line 375: `Math.round(numKb)` is turned into a integer, why it is casted by `toFixed(1)` to make a number look like `xxx.0`, why leave `.0` with one decimal but 0? 
If we like to have one decimal, it should be done by `(Math.round(numKb*10)/10).toFixed(1)` or just `String(Math.round(numKb*10)/10)`

packages/dropzone/src/dropzone.tsx
```diff
@@ -371,8 +371,8 @@ function useDropzoneStyles({
...
   }
-  return `${Math.round(numKb / 1024).toFixed(1)}MB`;
+  return `${Math.round(numKb / 1000).toFixed(1)}MB`;
```
freemany, line 377: Same here:
`(Math.round(numKb/100)/10).toFixed(1)`

packages/dropzone/src/dropzone.tsx
lukebennett88: Good call out Freeman.
When Danny and I built this, we where planning to use [this library](https://github.com/sindresorhus/pretty-bytes) to handle the math/formatting for us, but Next.js didn't like that it was ESM only.
We ended up slapping this together, planning to revisited it but never did.

ChristopherMitchell242: Shall I remove the toFixed(1) as part of this change? I do agree, i noticed it when i was doing some testing.... it looks a little odd.

lukebennett88 (approved):
LGTM 👍
Ship it! 🚢

PR #67, opened 2022-05-04
Add loading state to button

lukebennett88:


packages/button/src/Button.tsx
```diff
@@ -16,13 +20,25 @@ import { useButtonStyles } from './useButtonStyles';
...
+   * are controlled by the current element.
+   */
   'aria-controls'?: NativeButtonProps['aria-controls'];
+  /**  Identifies the element (or elements) that describes the object. */
   'aria-describedby'?: NativeButtonProps['aria-describedby'];
+  /**  Indicates whether the element, or another grouping element it controls, is currently expanded or collapsed. */
   'aria-expanded'?: NativeButtonProps['aria-expanded'];
+  /** When true, prevents onClick from firing. */
+  disabled?: boolean;
+  /** When true, the button will display a loading spinner. */
```
justinfaynhan, line 34: **Thought**: it would be amazing if these doc comments can be the source of truth instead of having these comments duplicated across with the README as well. When I last poked around options for this, there didn't seem to be an easy and customisable way unless we write our own parser...

justinfaynhan (approved):
LGTM

packages/button/src/Button.tsx
lukebennett88: Agreed. I've actually got a ticket to explore this:
https://brighte.atlassian.net/browse/SPRK-63

I found this package from Microsoft that we might be able to use:
https://www.npmjs.com/package/@microsoft/api-extractor

Will probably require quite a bit of config to get set up though.

lukebennett88: Oh wait, I just took a poke at the React Select docs to see what they are doing and looks like Atlassian have some packages to help with this: https://github.com/atlassian/extract-react-types

PR #68, opened 2022-05-04
Update lint-staged config

lukebennett88:
It doesn't look like `lint-staged` was working.
I've set it up from scratch, and also moved the config into the `package.json`. Seems to be working now 👍


package.json
```diff
@@ -22,13 +22,22 @@
...
     "lint:fix": "yarn run lint --fix",
+    "prepare": "husky install",
     "release": "yarn build && changeset publish",
     "start": "yarn --cwd=docs dev",
     "test": "yarn run typecheck && yarn run test:unit",
     "test:coverage": "yarn test ---coverage",
     "test:unit": "jest",
     "typecheck": "tsc --noEmit && cd docs && yarn typecheck"
   },
+  "lint-staged": {
```
justinfaynhan, line 33: suggestion(non-blocking): I prefer having lint-staged configurations in a `.lintstagedrc` file, better separation of concern and responsibilities.

justinfaynhan:
question: what exactly wasn't working with the initial setup? From what I tested, seemed to work for my workflows, was it due to not providing the husky install script?

lukebennett88:
> question: what exactly wasn't working with the initial setup? From what I tested, seemed to work for my workflows, was it due to not providing the husky install script?

I'm pretty sure the only thing missing was the `prepare` script, but I made a few small changes:

- Moved the lint-staged config into `package.json`. This is a personal preference, but I like to keep as much config as possible in package.json so the root doesn't explode with files! Whenever I check out a new repo, the first thing I do is look at the root `package.json`. [The co-founder of npm seems to recommend this](https://twitter.com/seldo/status/1500348639406821376) (but there are definitely good arguments on both sides).
- removed `yarn` from the beginning of commands to run in `lint-staged` as this is how the examples in the docs are formated.
- Added `.gitignore` in `.husky/` directory (as the contents of `.husky/_` are generated when the `prepare` script runs).

justinfaynhan (approved):

PR #71, opened 2022-05-05
Update React imports

lukebennett88:
When Spark Web was in the energy monorepo, we had to use the Babel and ESLint configs that were already there. Neither was configured to let you omit the React import for hinting to Babel/SWC to use `React.createElement` on JSX.

Now that we've got a bit more control this has been fixed. We did a big pass of the codebase when we pulled it out so that we could tree-shake React code from the packages, but missed a few places. This is just finishing that work off.

### DIFF
--- a/docs/components/mdx-components/mdx-table.tsx
+++ b/docs/components/mdx-components/mdx-table.tsx
@@ -3,13 +3,13 @@ import { Box } from '@spark-web/box';
 import { useHeading } from '@spark-web/heading';
 import { Text } from '@spark-web/text';
 import { useTheme } from '@spark-web/theme';
-import * as React from 'react';
+import type { ReactNode } from 'react';
 
 export function MdxTable({
   children,
   ...rest
 }: {
-  children: React.ReactNode;
+  children: ReactNode;
 }): JSX.Element {
   const theme = useTheme();
   return (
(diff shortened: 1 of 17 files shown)


docs/components/mdx-components/mdx-table.tsx
```diff
@@ -3,7 +3,6 @@ import { Box } from '@spark-web/box';
 import { useHeading } from '@spark-web/heading';
 import { Text } from '@spark-web/text';
 import { useTheme } from '@spark-web/theme';
-import * as React from 'react';
```
freemany, line 6: Can I ask what is the reason behind to add this react import in the first place, it is just recommended but I find that it can be skipped. cc @lukebennett88

freemany (approved):

docs/components/mdx-components/mdx-table.tsx
jesstelford: Next.js automatically adds this import for older versions of React. In newer versions, I don't think it's necessary at all anyway, so it can be omitted 👍

lukebennett88: I've updated the description to answer this, but the TLDR is ESLint and Babel.

PR #72, opened 2022-05-06
Feature/search optimisation

matildaPan:
Search PR: 

- Dev mode supports search index generation on hot reload , alone with content layer .
- Replace flexsearch with Lunr
- Refine Search async with React Suspend
- Search input box styles


docs/components/header.tsx
```diff
@@ -154,45 +154,104 @@ const GitHubLink = () => {
...
+
+const useSearch = (query: string) => {
+  if (!lunrIndex) {
+    throw getSearchInstance();
+  }
+  // Search with a post-fix wildcard, and fuzzy search (for minor typos)
+  return lunrIndex.search(`${query}*~1`);
+};
+
+const SearchResultsContainer = ({
```
matildaPan, line 189: we will replace SearchResultsContainer with Autocomplete box later

docs/next.config.mjs
```diff
@@ -1,92 +1,108 @@
...
 
-  allPackages.forEach(pkg => {
-    index.add({
-      slug: pkg.slug,
-      content: pkg.body.raw,
-    });
-  });
+  const index = lunr(function () {
+    this.ref('slug');
+    this.field('title', { boost: 100 });
```
matildaPan, line 25: based on our markdown , title takes more weight

docs/components/header.tsx
```diff
@@ -154,45 +154,103 @@ const GitHubLink = () => {
...
-        document: 'content',
-      });
-      flexsearchIndex.default.forEach(async (element: any) => {
-        await flexsearchDoc.import(element.key, JSON.parse(element.data));
+const getSearchInstance = async () => {
+  if (lunrIndex) {
+    return lunrIndex;
+  }
+  searchIndexPromise =
+    searchIndexPromise ??
```
freemany, line 165: I think it won't reach here line 165 but early return in 162.
we can try:
`if (lunrIndex) {`
`   return lunrIndex;`
`}`
`try {`
    `    const [jsonIndex, lunr] = await Promise.all([import('../cache/search-index.json'), import('lunr')]);`
    `   lunrIndex = lunr.default.Index.load(jsonIndex);`
`} catch (error) {`
    `   console.error(error);`
`}`
`return lunrIndex;`

matildaPan: It will reach here. At the first,  searchIndexPromise is `undefined`, so searchIndexPromise = Promis.all(search-index, lunr). On the second time, searchIndexPromise is unresolved promise, so it won't try to import async, hence we optimise the import for only just once.
Your code will work as well, but the downside is it may try to import more than once

.changeset/clean-eels-peel.md
```diff
@@ -0,0 +1,6 @@
+---
+'@spark-web/docs': minor
```
nderkim, line 2: @lukebennett88 said we don't version docs

docs/package.json
```diff
@@ -59,7 +59,11 @@
     "@untitled-docs/live-code": "^0.4.0",
     "autoprefixer": "^10.4.0",
     "clipboard-copy": "^4.0.1",
-    "flexsearch": "https://github.com/nextapps-de/flexsearch#bffb255b7904cb7f79f027faeb963ecef0a85dba",
+    "contentlayer": "^0.2.4",
```
nderkim, line 62: should these be in devDependencies?

docs/package.json
```diff
@@ -59,7 +59,11 @@
...
+    "esbuild": "^0.14.38",
```
nderkim, line 63: why is esbuild here?

docs/package.json
```diff
@@ -59,7 +59,11 @@
...
+    "gray-matter": "^4.0.3",
+    "lunr": "^2.3.9",
+    "mdx-bundler": "^9.0.0",
```
nderkim, line 66: same here

docs/package.json
lukebennett88: Yep, if we rebase from main that should already be fixed (although I spotted a few other dependencies that should probably be in devDependencies instead)

docs/package.json
lukebennett88: MDX Bundler needs it, but turns out we didn't need that either (Contentlayer comes with something for this, but just required some config to get working).
I've already fixed this in a PR yesterday.

packages/box/src/useBoxStyles.ts
```diff
@@ -167,7 +167,7 @@ export type ResponsiveBoxProps = {
    * The `top` property participates in specifying the vertical position of a
    * positioned element. It has no effect on non-positioned elements.
    */
-  top?: ResponsiveProp<0>;
+  top?: ResponsiveProp<0 | 100>;
```
nderkim, line 170: @lukebennett88 ☝️

lukebennett88: Yeah we should be using css function from Emotion instead of extending the accepted props.

nderkim (approved):

lukebennett88 (changes requested):
We're going to need a working Autocomplete before we can merge this.

docs/components/header.tsx
```diff
@@ -154,45 +154,103 @@ const GitHubLink = () => {
   );
 };
 
-let flexsearchDoc: any;
+let lunrIndex: any;
```
lukebennett88, line 157: Any chance we can get rid of this `any`.

docs/components/header.tsx
```diff
@@ -154,45 +154,103 @@ const GitHubLink = () => {
...
-  // );
+
+  lunrIndex = await searchIndexPromise;
+
+  return lunrIndex;
+};
+
+const useSearch = (query: string) => {
+  if (!lunrIndex) {
+    throw getSearchInstance();
```
freemany, line 182: Is this thrown error got caught?

docs/components/header.tsx
```diff
@@ -154,45 +154,103 @@ const GitHubLink = () => {
...
+        //@ts-expect-error
+        if (match?.[1].title) {
+          return (
+            <Text>
+              <TextLink href={`/package/${result.ref}`}>
+                Component &gt; <strong>{result.ref}</strong>
+              </TextLink>
+            </Text>
+          );
+        } else {
```
freemany, line 219: `else` can be removed

docs/components/header.tsx
matildaPan: it is React Suspense syntax

docs/components/header.tsx
matildaPan: no it can't be removed , it renders different UI component

docs/components/header.tsx
lukebennett88: Yep, this is how Suspense works.
I'm not an expert from what I know you `throw` a Promise and React will render the fallback in the suspense boundary. Once the Promise has resolved the whole component renders.

docs/components/header.tsx
freemany: I meant:
`if (match?.[1].title) {
          return ...
  }`
  `return ...;`

docs/components/header.tsx
```diff
@@ -2,31 +2,19 @@ import { css } from '@emotion/css';
...
-// @ts-expect-error flexsearch sucks
-import { Document as FlexSearchDocument } from 'flexsearch';
-import { useEffect } from 'react';
 
 import { GITHUB_URL, HEADER_HEIGHT, SIDEBAR_WIDTH } from './constants';
+import { Search } from './search';
 import { useSidebarContext } from './sidebar';
 import { BrighteLogo, GitHubIcon } from './vectors/fill';
 
 export function Header() {
```
lukebennett88, line 15: I've broken the Header up into smaller components to make more readable

lukebennett88 (approved):

PR #75, opened 2022-05-09
Sprk 56 build data attributes util for all components

ChristopherMitchell242:
The aim of SPRK-56 is to allow "most" (if not all) components to pass in a "data" attribute that can be spread into the component in question.


packages/box/src/Box.test.tsx
```diff
@@ -0,0 +1,30 @@
+import { cleanup, render } from '@testing-library/react';
+
+import { Box } from './Box';
+
+describe('Alert component', () => {
+  afterEach(cleanup);
+
+  it('should render correctly with the minimum set of props, without crashing', () => {
+    const { container } = render(<Box />);
```
freemany, line 9: Maybe we can have a `renderComponent` function like other tests:
```
const renderComponent = (props) => {
    return render('<Box ...props/>');
}
```

matildaPan (approved):
Awesome 🎉

freemany (approved):

PR #83, opened 2022-05-10
Feature/sprk 56 components data attr

ChristopherMitchell242:
Data props added to a load more components. Thanks in advance for the reviews!!

### DIFF
--- a/packages/checkbox/src/types.ts
+++ b/packages/checkbox/src/types.ts
@@ -1,4 +1,5 @@
 import type { Tone } from '@spark-web/field';
+import type { DataAttributeMap } from '@spark-web/utils/src/internal';
 import type { InputHTMLAttributes, ReactNode } from 'react';
 
 export type CheckboxSize = 'small' | 'medium';
@@ -11,6 +12,8 @@ type InputProps = Omit<
 export type CheckboxPrimitiveProps = {
   /** When true, the checkbox will be checked. */
   checked?: boolean;
+  /** Sets data attributes on the component. */
+  data?: DataAttributeMap;
   /** When true, the checkbox will be disabled. */
   disabled?: boolean;
   /** The size of the checkbox. */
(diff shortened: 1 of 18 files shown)


packages/checkbox/src/types.ts
```diff
@@ -26,4 +29,6 @@ export type CheckboxProps = {
   tone?: Tone;
   /** The checkbox label content. */
   children: ReactNode;
+  /** Sets data attributes on the component. */
+  data?: DataAttributeMap;
```
lukebennett88, line 33: `data` is doubled up here since this `CheckboxProps` has all of the props from `CheckboxPrimitiveProps`

packages/fieldset/src/fieldset.tsx
```diff
@@ -2,12 +2,15 @@ import { css } from '@emotion/css';
 import { Box } from '@spark-web/box';
 import { Text } from '@spark-web/text';
 import { useTheme } from '@spark-web/theme';
+import type { DataAttributeMap } from '@spark-web/utils/src/internal';
```
lukebennett88, line 5: Should be: `import type { DataAttributeMap } from '@spark-web/utils/internal';` as we don't ship the `src` directory to npm as of this PR: https://github.com/brighte-labs/spark-web/pull/52

lukebennett88: Just did a global search and this import has popped up a few times

ChristopherMitchell242: Ah nice, auto import sneaking that one in. I will do a global search and fix.

packages/checkbox/src/types.ts
ChristopherMitchell242: Thanks for the great review. I missed that!

lukebennett88 (approved):
LGTM 👍

PR #87, opened 2022-05-12
Feature/sprk 56 final component data attributes

ChristopherMitchell242:
Final set of components updated to include data attributes. Following the same patterns as previous commits.

### DIFF
--- a/.changeset/nine-ladybugs-stare.md
+++ b/.changeset/nine-ladybugs-stare.md
@@ -1,0 +1,19 @@
+---
+'@spark-web/inline': patch
+'@spark-web/link': patch
+'@spark-web/modal-dialog': patch
+'@spark-web/nav-link': patch
+'@spark-web/radio': patch
+'@spark-web/row': patch
+'@spark-web/select': patch
+'@spark-web/spinner': patch
+'@spark-web/stack': patch
+'@spark-web/text': patch
+'@spark-web/text-area': patch
+'@spark-web/text-input': patch
+'@spark-web/text-link': patch
+'@spark-web/text-list': patch
+---
+
+Data attribute added to component props to allow consumers to attach
+customattributes to components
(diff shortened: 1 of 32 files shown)


oscargws (approved):

packages/currency-input/src/CurrencyInput.tsx
```diff
@@ -15,7 +15,7 @@ import { currencySymbolMap } from './currencySymbolMap';
 /** NOTE: The "start" adornment is filled, so we can only accept a single child element. */
 export type CurrencyInputProps = {
   currencyType?: CurrencyType;
-  data: DataAttributeMap;
+  data?: DataAttributeMap;
```
lukebennett88, line 18: Good catch

packages/link/src/Link.test.tsx
```diff
@@ -0,0 +1,30 @@
+import { cleanup, render } from '@testing-library/react';
+
+import type { LinkProps } from './Link';
+import { Link } from './Link';
+
+describe('Link component', () => {
+  afterEach(cleanup);
+
+  const renderComponent = (props?: LinkProps) => {
+    const href = props?.href ? props?.href : 'test';
```
lukebennett88, line 10: Smallest nitpick ever: Second bit of optional chaining isn't required, can simply be:
`const href = props?.href ? props.href : 'test';`

packages/spinner/src/Spinner.test.tsx
```diff
@@ -2,6 +2,7 @@ import '@testing-library/jest-dom';
 
 import { render } from '@testing-library/react';
 
+import type { SpinnerProps } from '.';
```
lukebennett88, line 5: Don't think it matters in a test file that we're not exporting anything from, but as a rule of thumb I try to avoid importing from `.` since it's easy to accidentally end up with circular dependencies.

.changeset/nine-ladybugs-stare.md
```diff
@@ -0,0 +1,19 @@
+---
+'@spark-web/inline': patch
```
lukebennett88, line 2: `@spark-web/currency-input` needs a patch.

packages/link/src/Link.test.tsx
ChristopherMitchell242: Ha nice, good one!

.changeset/nine-ladybugs-stare.md
ChristopherMitchell242: Thanks!

PR #88, opened 2022-05-12
SPRK-38 Add TOC observer

nderkim:


docs/components/content/index.tsx
```diff
@@ -158,14 +159,22 @@ const HeadingList = () => {
...
+}: HeadingItemProps) => {
   const { color } = useTheme();
 
+  const isActive = useIsActive(id);
+
   const textStyles = useText({
     tone: 'neutral',
     size: level === 4 ? 'xsmall' : level === 3 ? 'small' : 'standard',
-    weight: 'regular',
+    weight: isActive ? 'strong' : 'regular',
```
lukebennett88, line 177: Rather than changing the weight of the active heading, maybe we should change the tone instead?
What do you think @jordangeizer?
https://spark-web-docs-git-alex-toc-observer-brighte.vercel.app/package/combobox#appearance

docs/components/content/intersection-observer.tsx
```diff
@@ -0,0 +1,96 @@
...
+      'useObserve must be used within a IntersectionObserverProvider'
+    );
+  }
+  const { observe } = context;
+  const ref = useRef<T>(null);
+  useEffect(
+    () => (ref.current ? observe?.(ref.current) : undefined),
+    [observe, ref]
+  );
+  return { ref };
```
lukebennett88, line 85: Nitpick/question: why do we return an object with a `ref` property in `useObserve`, but in `useEntries` we return the entries directly?

docs/components/content/toc-context.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+export const useTocContext = (): {
+  headings: HeadingData[];
+} => {
+  const context = useContext(TocContext);
+  if (typeof context === 'undefined') {
+    throw new Error('useTocContext must be used within a TocContextProvider');
+  }
+  return { headings: context };
+};
```
lukebennett88, line 32: Suggestion, rename `useTocContext` to `useHeadings` and create a generic hook that throws for us if context is undefined to save duplicate code:
```suggestion
export const useTocContext = () => {
  const context = useContext(TocContext);
  if (typeof context === 'undefined') {
    throw new Error('useTocContext must be used within a TocContextProvider');
  }
  return context;
};

export const useHeadings = () => {
  const headings = useTocContext();
  return { headings };
};
```

docs/components/content/intersection-observer.tsx
nderkim: `useEntries` returns `entries`
`useObserve` !returns `observe`

i could have done:
`useRef` returns `ref`
but then (aside from the fact that "useRef" is already taken) "useRef" doesn't explain the purpose of the hook anymore..

in my head i did:
```
const observe = { ref };
return observe;
```
and so
`useObserve` returns `observe`  :white_check_mark:

i feel like i just opened up my inner thoughts to a psychologist 😅

lukebennett88: But how does it make you _feel_? 😉
Thanks for that, makes perfect sense now 👍

docs/components/content/toc-context.tsx
nderkim: but then the next function will become:
```
export const Heading = (props: HeadingProps) => {
  useTocContext();
  ...
};
```
which doesn't actually use the context, and only calls the hook to throw the error, but someone else might think that it's not being used and delete it.
Also, the error message will say `useTocContext must be used within a TocContextProvider` as opposed to `Heading must be used within a TocContextProvider`

lukebennett88: Couldn't we rely on `useObserve` throwing the error?
That way we'd only need:
```tsx
export const Heading = (props: HeadingProps) => {
  const { ref } = useObserve<HTMLHeadingElement>();
  return <SparkHeading ref={ref} {...props} />;
};
```

lukebennett88 (approved):

docs/components/content/toc-context.tsx
nderkim: it does throw an error, but the error message will be `useObserve must be used within a IntersectionObserverProvider` as opposed to `Heading must be used within a TocContextProvider`

lukebennett88: Maybe the clearer error messages are worth the extra lines of code, IDK ¯\_(ツ)_/¯ 
PR looks good, I've approved so I'll let you decide 😊

lukebennett88: > it does throw an error, but the error message will be useObserve must be used within a IntersectionObserverProvider as opposed to Heading must be used within a TocContextProvider
🤦‍♂️ very good, carry on, nothing to see here.

nderkim: 🙈

PR #90, opened 2022-05-13
Feat/password input component

justinfaynhan:
Implemented new password input component.

### DIFF
--- a/packages/password-input/README.md
+++ b/packages/password-input/README.md
@@ -1,0 +1,86 @@
+---
+title: Password Input
+storybookPath: forms-passwordinput--default
+---
+
+The password input component allows accepting of password text values. The
+component builds off its underlying [TextInput](/package/text-input) component
+and toggles the input mode between `password` and `text` depending on whether
+the show password icon button is toggled to show or hide the password value.
+
+## Usage
+
+The component must be nested within a [`Field`](/package/field). See
+[`Field`](/package/field) for more details.
+
+## Examples
+
+### Field utilities
+
+You can use the parent [Field](/package/field) to set appropriate field level
+labels, description adornments and messages.
+
+```jsx live
+<Stack gap="large">
+  <Field
+    label="Password"
+    adornment={
+      <Text>
+        <TextLink href="#">Forgot password?</TextLink>
+      </Text>
+    }
+    description={
+      'Make sure your chosen password is at least 10 characters long.'
+    }
+  >
+    <PasswordInput />
+  </Field>
+</Stack>
+```
+
+### Uncontrolled
+
+Similar to [TextInput](/package/text-input), the component can be uncontrolled
+and managing its own internal state.
+
+```jsx live
+const passwordRef = React.useRef(null);
+const [value, setValue] = React.useState('');
+const showValueHandler = React.useCallback(() => {
+  setValue(passwordRef.current?.value);
+}, [setValue]);
+
+return (
+  <Stack gap="large">
+    <Field label="Example uncontrolled">
+      <PasswordInput ref={passwordRef} />
+    </Field>
+    <Button onClick={showValueHandler}>Display password value</Button>
+    <Text>The password value is: {value}</Text>
+  </Stack>
+);
+```
+
+### Controlled
+
+Similar to [TextInput](/package/text-input), the component can be controlled by
+passing in a `value` and `onChange` handler and have its state managed in a
+parent component.
+
+```jsx live
+const [value, setValue] = React.useState(``);
+
+return (
+  <Stack gap="large">
+    <Field label="Example controlled">
+      <PasswordInput value={value} onChange={e => setValue(e.target.value)} />
+    </Field>
+    <Text>The current password value is: {value}</Text>
+  </Stack>
+);
+```
+
+## Props
+
+The component passes props into its underlying [TextInput](/package/text-input)
+component and are not listed here.
--- a/packages/password-input/README.md
+++ b/packages/password-input/README.md
@@ -82,5 +82,9 @@ return (
 
 ## Props
 
+| Prop        | Type   | Default | Description                                                    |
+| ----------- | ------ | ------- | -------------------------------------------------------------- |
+| aria-label? | string |         | Sets an accessibility label for the show password icon button. |
+
 The component passes props into its underlying [TextInput](/package/text-input)
 component and are not listed here.
--- a/packages/password-input/README.md
+++ b/packages/password-input/README.md
@@ -82,9 +82,5 @@ return (
 
 ## Props
 
-| Prop        | Type   | Default | Description                                                    |
-| ----------- | ------ | ------- | -------------------------------------------------------------- |
-| aria-label? | string |         | Sets an accessibility label for the show password icon button. |
-
 The component passes props into its underlying [TextInput](/package/text-input)
 component and are not listed here.
--- a/packages/password-input/src/IconButton.tsx
+++ b/packages/password-input/src/IconButton.tsx
@@ -16,15 +16,8 @@ export const IconButton = ({
   ...rest
 }: IconButtonProps) => {
   return (
-    <>
-      <Box
-        as="button"
-        onClick={handleClick}
-        {...useIconButtonStyles()}
-        {...rest}
-      >
-        {children}
-      </Box>
-    </>
+    <Box as="button" onClick={handleClick} {...useIconButtonStyles()} {...rest}>
+      {children}
+    </Box>
   );
 };
--- a/packages/password-input/src/IconButton.tsx
+++ b/packages/password-input/src/IconButton.tsx
@@ -7,12 +7,23 @@ import { useIconButtonStyles } from './useIconButtonStyles';
 type IconButtonProps = {
   handleClick: NativeButtonProps['onClick'];
   children: ReactNode;
+  'aria-label'?: string;
+  'aria-pressed'?: boolean;
 };
 
-export const IconButton = ({ handleClick, children }: IconButtonProps) => {
+export const IconButton = ({
+  handleClick,
+  children,
+  ...rest
+}: IconButtonProps) => {
   return (
     <>
-      <Box as="button" onClick={handleClick} {...useIconButtonStyles()}>
+      <Box
+        as="button"
+        onClick={handleClick}
+        {...useIconButtonStyles()}
+        {...rest}
+      >
         {children}
       </Box>
     </>
--- a/packages/password-input/src/IconButton.tsx
+++ b/packages/password-input/src/IconButton.tsx
@@ -1,12 +1,12 @@
 import { Box } from '@spark-web/box';
-import type { ButtonHTMLAttributes, ReactNode } from 'react';
+import type { IconProps } from '@spark-web/icon';
+import type { ButtonHTMLAttributes, ReactElement } from 'react';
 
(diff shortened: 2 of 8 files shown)


packages/password-input/src/IconButton.tsx
```diff
@@ -0,0 +1,20 @@
+import { Box } from '@spark-web/box';
+import type { NativeButtonProps } from '@spark-web/button/src/types';
```
lukebennett88, line 2: We can't import from the `src` directory of our packages as that folder isn't published to npm.

justinfaynhan: Good catch, easy fix

packages/password-input/README.md
```diff
@@ -0,0 +1,90 @@
...
+    <Text>The current password value is: {value}</Text>
+  </Stack>
+);
+```
+
+## Props
+
+| Prop        | Type   | Default | Description                                                    |
+| ----------- | ------ | ------- | -------------------------------------------------------------- |
+| aria-label? | string |         | Sets an accessibility label for the show password icon button. |
```
justinfaynhan, line 87: @lukebennett88 is it a good idea to publicly expose aria-* props as part of the API, or should this be extracted away and handled under the hood in the component

packages/password-input/src/PasswordInput.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { useFieldContext } from '@spark-web/field';
+import { EyeIcon, EyeOffIcon } from '@spark-web/icon';
+import type { TextInputProps } from '@spark-web/text-input';
+import { InputAdornment, TextInput } from '@spark-web/text-input';
+import { forwardRef, useState } from 'react';
+
+import { IconButton } from './IconButton';
+
+export type PasswordInputProps = {
+  'aria-label'?: string;
```
justinfaynhan, line 11: @lukebennett88 is it a good idea to publicly expose aria-* props as part of the API, or should this be extracted away and handled under the hood in the component

packages/password-input/src/IconButton.tsx
```diff
@@ -0,0 +1,30 @@
+import { Box } from '@spark-web/box';
+import type { ButtonHTMLAttributes, ReactNode } from 'react';
+
+import { useIconButtonStyles } from './useIconButtonStyles';
+
+type IconButtonProps = {
+  handleClick: ButtonHTMLAttributes<HTMLButtonElement>['onClick'];
+  children: ReactNode;
```
lukebennett88, line 8: We can tighten this type up to only accept our icons as children:
```tsx
children: ReactElement<IconProps>;
```

packages/password-input/src/IconButton.tsx
```diff
@@ -0,0 +1,30 @@
...
+  'aria-label'?: string;
```
lukebennett88, line 9: `aria-label` should be required for this button since it should only take an icon as children.

packages/password-input/src/PasswordInput.tsx
lukebennett88: It's not clear that `aria-label` is forwarded onto the IconButton. I'm also not sure that consumers should need to be able to change this label. I'd remove this prop and hard-code the label on the icon button:
```tsx
<IconButton
  aria-label="Toggle password visibility"
  aria-pressed={showPassword}
  handleClick={handleClick}
>
```

packages/password-input/src/PasswordInput.tsx
```diff
@@ -0,0 +1,50 @@
...
+        inputMode={'text'}
+        type={showPassword ? 'text' : 'password'}
+      >
+        <InputAdornment placement="end">
+          <IconButton
+            handleClick={handleClick}
+            aria-label={ariaLabel}
+            aria-pressed={showPassword}
+          >
+            {showPassword ? (
```
lukebennett88, line 38: Nitpick: The ternary and duplication of props could be made a little more readable like this:

```tsx
// on line 24
const Icon = showPassword ? EyeOffIcon : EyeIcon;
```

```tsx
// on line 38
<Icon tone={disabled ? 'disabled' : 'neutral'} />
```

packages/password-input/src/PasswordInput.tsx
```diff
@@ -0,0 +1,50 @@
...
+    const handleClick = getPreventableClickHandler(
+      toggleShowPassword,
+      disabled
+    );
+
+    return (
+      <TextInput
+        {...props}
+        ref={forwardedRef}
+        inputMode={'text'}
```
lukebennett88, line 29: Nitpick: we don't need to curly braces around `'text'`.

packages/password-input/src/useIconButtonStyles.ts
```diff
@@ -0,0 +1,36 @@
...
+    // interactions styles
+    className: css({
+      '&:hover': {
+        backgroundColor: theme.backgroundInteractions['neutralLowHover'],
+      },
+      '&:active': {
+        backgroundColor: theme.backgroundInteractions['neutralLowActive'],
+        transform: 'scale(0.98)',
+      },
+      ':focus': focusRingStyles,
```
lukebennett88, line 31: @jordangeizer what should we do regarding hover/active/focus styles for this button?
[Here is a playroom link to try it out.](https://spark-web-docs-8xgtc26iq-brighte.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8BlALgQzAawAQHMMAHAXgB0QAbDAJ3xgtyIyigEsA7fcq2%2BigHxkOuXEgBibGJSjDRo6gCNpPAAoYAzhoDuEGrJBz5LPRwC2MDmhLAj8sQBUYADzRCR9%2BUieuAMpzwACxoYADMeAGJBcT18CDQmTR09KAB%2BJAB6HzR-Dmx3TzEslzc7XABfMtgNMBo2IjQ2CA4bMtEKAFkMbBhcDQBXENwATwhB3DBAiA1LRK1dfVw2DVwMBMoYTQSARgAGCcDaLDQYGhXKZvwAOgoyyo8Cr3V5lIBJDiJ%2BhIzHzMlpKDuTLoLD5YQgcpAA)

lukebennett88: A tooltip might be a nice touch for this component 🤔

packages/password-input/src/IconButton.tsx
justinfaynhan: Removed the aria-label prop

packages/password-input/src/PasswordInput.tsx
justinfaynhan: Removed the aria-label prop

packages/password-input/src/PasswordInput.tsx
```diff
@@ -0,0 +1,42 @@
+import { getPreventableClickHandler } from '@spark-web/button/src/Button';
```
lukebennett88, line 1: We can't import from `src` directories of packages. 
Will need to export this function from the index file create a patch changeset.

justinfaynhan: I'm leaning towards copy pasting the getPreventableClickHandler into PasswordInput instead.
✅ Removes extra dependency of passwordinput on button
✅ getPreventableClickHandler isn't too sophisticated
✅ Keeps Button exposed API surface cleaner
Yes there is duplicated code, but we can consolidate once we decide icon button is worth extracting out into its own seperate component.

justinfaynhan: Also will be worth adding eslint rules to prohibit `@spark-web/**/src/**` imports (see https://eslint.org/docs/rules/no-restricted-imports)

lukebennett88: Great suggestions ✅

packages/password-input/src/useIconButtonStyles.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  const buttonStyleProps: Partial<BoxProps> = {
+    alignItems: 'center',
+    borderRadius: 'full',
+    cursor: 'pointer',
+    display: 'inline-flex',
+    gap: 'small',
+    height: 'small',
+    justifyContent: 'center',
+    marginTop: 'small',
```
lukebennett88, line 18: We don't want any margin on this.
If you rebase from `main` it should be centered automatically now:
https://github.com/brighte-labs/spark-web/pull/94

packages/password-input/src/IconButton.tsx
```diff
@@ -0,0 +1,30 @@
...
+  'aria-pressed'?: boolean;
+};
+
+export const IconButton = ({
+  handleClick,
+  children,
+  ...rest
+}: IconButtonProps) => {
+  return (
+    <>
```
freemany, line 19: This fragment is redundant.

packages/password-input/src/useIconButtonStyles.ts
jordangeizer: Thanks @lukebennett88! Agreed! Tooltips would be nice here. I'll make a note of that :D 

It's looking great! Some thoughts I had:
- Not sure if your other PR will fix this too - but I realised we seem to loose the focus appearance on hover of the button (despite the button still being focussed).
- We probably don't have a token for it, but I was wondering if the hover outline for the button could be a neutral hue to match the icon/background appearance?
- Since the button is much smaller than the others, I realise we might want to drop that scale down to: `transform: 'scale(0.94)',`

lukebennett88: > Not sure if your other PR will fix this too - but I realised we seem to loose the focus appearance on hover of the button (despite the button still being focussed).

I've noticed this too, I think it's expected behaviour for the focus-visible hook. As far as I know, the idea is; if you focus elements with your keyboard you want a focus ring (so you can see what has focus) but if you do it with your mouse, chances are you _don't_ want a focus ring. `useFocusVisible` has an event listener listening for a mouse move event, and when that happens it removes the focus ring.

> We probably don't have a token for it, but I was wondering if the hover outline for the button could be a neutral hue to match the icon/background appearance?

We might patch this in a future PR if that's OK. We're actually passing the `neutral` tone to `useFocusRing`, but that tone creates a green focus ring for our neutral buttons. [See Playroom link](https://spark.brighte.com.au/playroom#?code=N4Igxg9gJgpiBcIA8BlALgQzAawAQHMMAHAXgB0QAbDAJ3xgtyIyigEsA7fcq2%2BigHxkOuXEgCSHSpxgFiPAM4BbDJUqDho0UgBCAVzRoIIoxxg8zBmqsZEaEJTI5hzFABZt8bjSK2iAcjBWNr7aAPT6hsZCoWKRprimriCWaNbqIEz2jmbOyZQQAO4%2BfgFBaSF%2BSBEGpjFV8caJxsmp6bbZTi4WLSWlgcHqsdWNHPViYZLSZjHV6FjYMSAAvkA).

> Since the button is much smaller than the others, I realise we might want to drop that scale down to: transform: 'scale(0.94)',

Good idea, can you also add some transitions so the scale transform is animated @justinfaynhan  🙏
```tsx
transitionProperty:
  'color, background-color, border-color, text-decoration-color, fill, stroke, opacity, box-shadow, transform, filter, backdrop-filter',
transitionTimingFunction: 'cubic-bezier(0.4, 0, 0.2, 1)',
transitionDuration: `${theme.animation.standard.duration}ms`,
```

lukebennett88 (approved):

PR #94, opened 2022-05-16
Fix alignment of icon adornments

lukebennett88:
Before:

PixelSnap 2022-05-16 at 10 30 57@2x

After:

PixelSnap 2022-05-16 at 10 31 33@2x


ChristopherMitchell242 (approved):
Eyes of a hawk

PR #96, opened 2022-05-16
feat(combobox): expose getOptionLabel and getOptionValue props

clothoo:


tamm:
You need to add a changeset. Also you haven't noted the change to aria-errormessage in the PR, that needs to be mentioned ^_^

packages/field/src/context.ts
```diff
@@ -2,6 +2,7 @@ import { createContext, useContext } from 'react';
 
 export type FieldContextType = {
   'aria-describedby'?: string;
+  'aria-errormessage'?: string;
```
lukebennett88, line 5: `aria-errormessage` doesn't actually ever get put into the context provider:
https://github.com/brighte-labs/spark-web/blob/main/packages/field/src/Field.tsx#L72-L80

Do you mind removing it for this PR? I'll start investing how we can improve the accessibility of the Field component and address it in a future PR.

packages/combobox/README.md
```diff
@@ -69,6 +69,33 @@ return (
...
+        { title: 'The Godfather: Part II', year: 1974 },
+        { title: 'The Dark Knight', year: 2008 },
+        { title: '12 Angry Men', year: 1957 },
+        { title: "Schindler's List", year: 1993 },
+        { title: 'Pulp Fiction', year: 1994 },
+      ]}
+      onChange={value => setValue(value)}
+      value={value}
+      getOptionLabel={option => option.title}
+      getOptionValue={option => option.year}
```
lukebennett88, line 93: I just tried this example locally and am getting a type error because React Select wants `year` to be a string instead of a number.

packages/field/src/context.ts
clothoo: Sure, I was hoping I can use it for combobox alongside field. Currently combobox `aria-describedby` is pointing to `placeholder`. The message from `Field` is not accessible.

lukebennett88: Good catch, React Select must be doing that. Will investigate further 👀

packages/combobox/src/combobox.tsx
```diff
@@ -25,6 +26,15 @@ export type ComboboxProps<Item = unknown> = {
...
   value?: Nullable<Item>;
+  /**
+   * Resolves option data to a string to be displayed as the label by components
+   *
+   * Note: Failure to resolve to a string type can interfere with filtering and
+   * screen reader support.
+   */
+  getOptionLabel?: GetOptionLabel<Item>;
+  /** Resolves option data to a string to compare options and specify value attributes */
+  getOptionValue?: GetOptionValue<Item>;
```
justinfaynhan, line 37: **Suggestion**: Can you also add these component docs in the README props table? Unfortunately, we don't have a good way of generating consumer-facing component props documentation from js-docs yet, so the comments and docs are duplicated. README documentation takes precedence at this stage.

.changeset/neat-tools-shout.md
```diff
@@ -0,0 +1,5 @@
+---
+'@spark-web/combobox': patch
```
lukebennett88, line 2: This should be a minor since there is new functionality

packages/combobox/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@spark-web/combobox",
-  "version": "1.0.0",
+  "version": "1.0.1",
```
lukebennett88, line 3: We don't need to bump this ourselves, [Changeset](https://github.com/changesets/changesets) handles this for us when we publish a release.

lukebennett88 (approved):
LGTM 👍
Thanks Shawn!

ChristopherMitchell242 (approved):
Great!

Thanks mate!

packages/combobox/README.md
```diff
@@ -69,6 +69,33 @@ return (
...
+    />
+  </Field>
+);
+```
+
```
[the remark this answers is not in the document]
lukebennett88: Should already be there:
PixelSnap 2022-05-17 at 12 28 17@2x

PR #99, opened 2022-05-17
Remove props spreading from spark components

oscargws:


packages/link/src/Link.tsx
```diff
@@ -6,8 +6,10 @@ import { useLinkComponent } from './linkContext';
 
 export type LinkProps = LinkComponentProps;
 
-export const Link = forwardRefWithAs<'a', LinkProps>((props, ref) => {
+export const Link = forwardRefWithAs<'a', LinkProps>(({ href, data }, ref) => {
```
lukebennett88, line 9: I reckon this component might be OK to keep to consumer prop spreading. It's just an un-styled link component, without the extra props it's reasonably useless!

lukebennett88: Also `LinkProps` has all the valid props for an anchor element, so you can pass `Link` a bunch of props, but it won't do anything with them so it fails silently.
e.g. in the deploy preview for this PR the Brighte and GitHub logos aren't styled since the `className` prop isn't getting forwarded on.

oscargws: Yeah was tweaking with this on the preview branch, was hoping I could get it working without the spread but think you're right

.changeset/red-flies-love.md
```diff
@@ -0,0 +1,16 @@
+---
+'@spark-web/a11y': major
```
ChristopherMitchell242, line 2: I dont see a change to the a11y component, can we remove this bump? Or am I missing something?

packages/row/src/Row.tsx
```diff
@@ -28,7 +28,7 @@ export type RowProps = {
 
 export const Row = forwardRefWithAs<'div', RowProps>(
   (
-    { align = 'left', alignY = 'stretch', children, data, dividers, ...props },
+    { align = 'left', alignY = 'stretch', children, data, dividers },
```
lukebennett88, line 31: We've still got the same problem as the we did with the `Columns`.
Because `ValidBoxProps` is removing a handful of props from `BoxProps`, everything that remains is still a valid prop, but we're not doing anything with them.
e.g. this component won't pass on any of the props we give it:
```tsx
function RowButton() {
  return (
    <Row as="button" aria-describedby="description" />
  );
};
```

oscargws: Yep, how do we want to handle this, just let the ...props stay for columns and boxes, or pass every single BoxProps option as a prop directly though, feels  bit clunky?

lukebennett88: I'm not sure.
Removing the `ValidBoxProps` doesn't work here since `forwardRefWithAs` pretty much lets through anything that is valid for that element.
I'm thinking we might have to put the spread back (making sure it's above everything else so it can't overwrite props that we're already using) and just add a few more items to the `Omit` (for instance `className`, `style`, `dangerouslySetInnerHTML`).

Curious if anyone has a better idea though. @jossmac, @JedWatson, @nderkim.

lukebennett88: If we _can_ get away with removing spreading of consumer props, we're going to need to add more props to this list.
`gap` is the big one, but probably `as` and a few others.

.changeset/red-flies-love.md
```diff
@@ -0,0 +1,14 @@
+---
+'@spark-web/combobox': major
+'@spark-web/currency-input': major
+'@spark-web/field': major
+'@spark-web/heading': major
+'@spark-web/link': major
+'@spark-web/modal-dialog': major
+'@spark-web/radio': major
+'@spark-web/row': major
+'@spark-web/text': major
```
ChristopherMitchell242, line 10: I dont see any change associated with the text component? Did you revert a change?

.changeset/red-flies-love.md
```diff
@@ -0,0 +1,14 @@
+---
+'@spark-web/combobox': major
```
ChristopherMitchell242, line 2: combobox change has been reverted? Should this be Columns?

packages/radio/src/radio-group.tsx
```diff
@@ -6,12 +6,24 @@ import { useRadioGroupState } from './use-radio-group-state';
 
 export const RadioGroup = <Value extends string>({
   children,
-  id: idProp,
+  disabled,
```
ChristopherMitchell242, line 9: Do we need to update the readme for these props now they are declared explicitly?

PR #101, opened 2022-05-18
Update focus and disabled styles for form elements

lukebennett88:
One thing that we considered when designing our inputs is that password managers like to insert their icons _inside_ of the inputs. This can overlap with adornments (buttons/icons etc) so we are careful to wrap our inputs in an element that wraps both the input and adornments and fakes the focus ring. This way the input sits beside the adornment instead of of the adornment being absolutely positioned above the input.

Previously we were using the :focus-within` pseudo-class to apply the focus ring to the wrapper div whenever any child had focus, however, this causes double focus rings when we focus on the adornment (for things like the [PasswordInput](https://github.com/brighte-labs/spark-web/pull/90)).

The way I've come up with to address this is to stop using `:focus-within` and instead style an absolutely positioned adjacent sibling whenever the input has focus. This does mean that we need to export a component that sits beside our inputs to be styled. I tried styling a `::before` pseudo element so we didn't need to extra element, but I couldn't get it to work. I'm not 100% sure why, but I think it's because inputs are treated as 'replaced' elements and so the `::before` and `::after` don't work on them.

Here's some screenshots of what it looks like in action:

Adornment and 1Password icon sitting beside each other
PixelSnap 2022-05-18 at 10 53 49@2x

Adornment focused, no focus ring on input
PixelSnap 2022-05-18 at 10 54 33@2x

### DIFF
--- a/.changeset/tiny-bananas-doubt.md
+++ b/.changeset/tiny-bananas-doubt.md
@@ -1,0 +1,5 @@
+---
+'@spark-web/text-input': minor
+---
+
+Create FocusIndicator component and update focus styles
--- a/.changeset/tiny-bananas-doubt.md
+++ b/.changeset/tiny-bananas-doubt.md
@@ -2,4 +2,4 @@
 '@spark-web/text-input': minor
 ---
 
-Create FocusIndicator component and update focus and disabled styles
+Create InputContainer component and update focus and disabled styles
--- a/.changeset/tiny-bananas-doubt.md
+++ b/.changeset/tiny-bananas-doubt.md
@@ -2,4 +2,4 @@
 '@spark-web/text-input': minor
 ---
 
-Create FocusIndicator component and update focus styles
+Create FocusIndicator component and update focus and disabled styles
--- a/packages/select/src/Select.tsx
+++ b/packages/select/src/Select.tsx
@@ -3,7 +3,7 @@ import { Box } from '@spark-web/box';
 import { useFieldContext } from '@spark-web/field';
 import { ChevronDownIcon } from '@spark-web/icon';
 import type { UseInputProps } from '@spark-web/text-input';
-import { FocusIndicator, useInput } from '@spark-web/text-input';
+import { InputContainer, useInput } from '@spark-web/text-input';
 import { useTheme } from '@spark-web/theme';
 import type { DataAttributeMap } from '@spark-web/utils/internal';
 import type { SelectHTMLAttributes } from 'react';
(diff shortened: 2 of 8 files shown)


packages/text-area/src/text-area.tsx
```diff
@@ -48,12 +49,16 @@ export const TextArea = forwardRef<HTMLTextAreaElement, TextAreaProps>(
...
+      <Box
+        position="relative"
+        background={disabled ? 'inputDisabled' : 'input'}
+      >
         <Box
           {...a11yProps}
           {...consumerProps}
-          data={data}
           as="textarea"
+          aria-invalid={invalid || undefined}
```
lukebennett88, line 60: I missed TextArea [in this PR](https://github.com/brighte-labs/spark-web/pull/98)

freemany: You only want `string`, `true` or `number` to be returned otherwise `undefined` ?

lukebennett88: `invalid` is an optional boolean.
If it's `true` I want it to be `aria-invalid={true}`. 
If it's `false` I want it to be `aria-invalid={undefined}` so the attribute doesn't show up in the DOM.

ChristopherMitchell242 (approved):

packages/select/src/Select.tsx
```diff
@@ -54,7 +54,19 @@ export const Select = forwardRef<HTMLSelectElement, SelectProps>(
     );
 
     return (
-      <Box position="relative">
+      <InputContainer>
+        <Box
```
lukebennett88, line 58: Had to move the chevron above the `<select>` so that the styles for the focus indicator would get applied.
Thankfully the select is `position: static` so there was no need to mess with z-index.

.changeset/tiny-bananas-doubt.md
```diff
@@ -0,0 +1,5 @@
+---
+'@spark-web/text-input': minor
+---
+
+Create FocusIndicator component and update focus and disabled styles
```
jossmac, line 5:
```suggestion
Create InputContainer component and update focus and disabled styles
```

packages/text-area/src/text-area.tsx
```diff
@@ -48,12 +49,13 @@ export const TextArea = forwardRef<HTMLTextAreaElement, TextAreaProps>(
...
 
     return (
-      <Box position="relative">
+      <InputContainer>
         <Box
           {...a11yProps}
           {...consumerProps}
-          data={data}
           as="textarea"
+          aria-invalid={invalid || undefined}
```
jossmac, line 57: For another PR: let's move this behaviour into field context.

packages/text-input/src/TextInput.tsx
```diff
@@ -120,28 +115,58 @@ export const useInput = ({ disabled, isNested = false }: UseInputProps) => {
...
+    ':enabled': {
+      ':focus + [data-focus-indicator]': {
+        borderColor: theme.border.color.fieldAccent,
+        ...focusRingStyles,
+      },
+    },
   } as const;
 };
+
+export type InputContainerProps = {
```
jossmac, line 128: For another PR: this probably deserves its own file.

jossmac (approved):
LGTM! 🚢

packages/text-area/src/text-area.tsx
jossmac: I wonder if it would be tidier to explicitly separate the concerns 🤔 Something like:

```ts
type FieldState = {
  disabled: boolean;
  invalid: boolean;
}
type InputPropsDerivedFromField = {
  'aria-describedby'?: string;
  'aria-invalid'?: boolean;
  id: string;
}
type FieldContextType = [FieldState, InputPropsDerivedFromField]
```

PR #104, opened 2022-05-18
Load fonts from Brighte subdomain

lukebennett88:


packages/core/src/fonts.tsx
```diff
@@ -1,12 +1,23 @@
...
+import { Global } from '@emotion/react';
+
 export function AesteticoStylesheet(): JSX.Element {
-  return (
-    <link
-      rel="stylesheet"
-      href="https://spark-web-docs-fa4i8lp4e-brighte.vercel.app/fonts.css"
-    />
-  );
+  return <Global styles={fonts} />;
```
lukebennett88, line 4: Keeping this in a seperate file so that consumers of the design system aren't forced to load the Aestetico font (e.g. if they want to override the theme).
It also means this isn't a breaking change.

ChristopherMitchell242 (approved):

PR #108, opened 2022-05-19
Refactor useFieldContext to return tuple

lukebennett88:
# Description

This PR addresses some feedback from Joss:
https://github.com/brighte-labs/spark-web/pull/101#discussion_r876582101
It refactors the `useFieldContext` hook to return a tuple of state and derived props and adds `aria-invalid`.

# Checklist:

### DIFF
--- a/packages/text-input/src/TextInput.tsx
+++ b/packages/text-input/src/TextInput.tsx
@@ -2,7 +2,7 @@ import { css } from '@emotion/css';
 import { useFocusRing } from '@spark-web/a11y';
 import type { BoxProps } from '@spark-web/box';
 import { Box } from '@spark-web/box';
-import type { FieldContextType } from '@spark-web/field';
+import type { FieldState } from '@spark-web/field';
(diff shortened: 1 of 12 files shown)


packages/text-input/src/TextInput.tsx
```diff
@@ -98,7 +97,7 @@ export const TextInput = forwardRef<HTMLInputElement, TextInputProps>(
 
 TextInput.displayName = 'TextInput';
 
-export type UseInputProps = Pick<FieldContextType, 'disabled' | 'invalid'>;
+export type UseInputProps = Pick<FieldContextType[0], 'disabled' | 'invalid'>;
```
jossmac, line 100: Is it worth exposing `FieldState` and `InputPropsDerivedFromField` for cases like this?

lukebennett88: lol, I didn't think about that one _at all_ — just blindly fixed the types. Good call, updated 👍

ChristopherMitchell242 (approved):

PR #116, opened 2022-05-20
Update Babel config for Playroom

lukebennett88:
# Description

Some of our Playroom links are broken. This is because we're using a different Babel config for our Playroom to the one we use to build our components.


matildaPan (approved):
LGTM

PR #118, opened 2022-05-22
SPRK-83 Combobox Data Attribute

ChristopherMitchell242:
# Description

Adding Data Attribute to Combobox component to bring this component in line with all the others. 

# Associated Tickets

- SPRK-83

# Checklist:

- [X] I've updated unit tests where needed
- [X] I've updated the components README.md where needed
- [X] I've added changesets where needed
- [X] All the components I've updated are reflected in the changelog

### DIFF
--- a/.changeset/beige-rockets-kick.md
+++ b/.changeset/beige-rockets-kick.md
@@ -1,0 +1,5 @@
+---
+'@spark-web/combobox': patch
+---
+
+Add data attribute to Combobox component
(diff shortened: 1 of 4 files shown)


lukebennett88:
This PR would benefit from a test to verify that the data-attribute is actually being forwarded on.
We have a ticket to [investigate why some aria attributes aren't being forwarded on](https://brighte.atlassian.net/browse/SPRK-84) for the Combobox, I wouldn't be surprised if React Select doesn't forward the data attribute on also.

PR #122, opened 2022-05-23
Control spacing of Fieldset children

lukebennett88:
# Description

Fieldset has a `gap` prop, but it's only used for the gap between the legend and the content.
This PR adds a Stack to control space between the `children`.
If users require different spacing between `children` for they can wrap their children in another `Stack`, but this should provide a sensible default.


oscargws (approved):
A premium DX change 🔥

PR #126, opened 2022-05-23
Load fonts from Brighte subdomain

lukebennett88:
# Description

Once the CORS issues are fixes, we can load fonts from the Brighte subdomain. Until then I'm leaving this PR open.

### DIFF
--- a/packages/core/src/fonts.tsx
+++ b/packages/core/src/fonts.tsx
@@ -1,30 +1,30 @@
 import { Global } from '@emotion/react';
 
-export function AesteticoStylesheet(): JSX.Element {
-  return <Global styles={fonts} />;
-}
+export const AESTETICO_REGULAR_URL =
+  'https://static-assets.prod.cloud.brighte.com.au/fonts/Aestetico-Regular.woff2';
+export const AESTETICO_SEMIBOLD_URL =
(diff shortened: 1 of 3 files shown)


packages/core/src/fonts.tsx
```diff
@@ -16,15 +16,13 @@ const fonts = `
...
+  src: url(https://static-assets.prod.cloud.brighte.com.au/fonts/Aestetico-Regular.woff2) format("woff2");
 }
 @font-face {
   font-family: Aestetico;
   font-style: normal;
   font-weight: 600;
   font-display: swap;
-  /* src: url(https://static-assets.prod.cloud.brighte.com.au/fonts/Aestetico-SemiBold.woff2) format("woff2"); */
-  src: url(https://spark-web-docs-fa4i8lp4e-brighte.vercel.app/fonts/aestetico-semibold.woff2) format("woff2");
+  src: url(https://static-assets.prod.cloud.brighte.com.au/fonts/Aestetico-SemiBold.woff2) format("woff2");
```
Thinkscape, line 26: My suggestion would be to extract the `baseUrl` into a config-like file/constant, to potentially allow customisation.

Might be out of scope of this changeset 🤷

Thinkscape (approved):

PR #127, opened 2022-05-23
Fix the `reserve-space` prop reserving the wrong amount of space

lukebennett88:
# Description

After merging [this PR](https://github.com/brighte-labs/spark-web/pull/123), I noticed that the `reserve-space` prop was now reserving the wrong amount of space. This PR fixes that.

Before:
PixelSnap 2022-05-23 at 16 47 05@2x

After:
PixelSnap 2022-05-23 at 16 47 23@2x

Changes are already covered by this changeset:
https://github.com/brighte-labs/spark-web/blob/main/.changeset/two-islands-fold.md

### DIFF
--- a/packages/field/src/Field.tsx
+++ b/packages/field/src/Field.tsx
@@ -110,9 +110,7 @@ export const Field = forwardRef<HTMLDivElement, FieldProps>(
       'reserve-space': (
         <Fragment>
           {hiddenLabel}
-          <Text inline aria-hidden>
-            &nbsp;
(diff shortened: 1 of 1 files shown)


jossmac (approved):

jossmac:
Related: We can probably remove the "reserveSpace" option from `labelVisibility` now that we have the `Fieldset`, which is a better [semantic solution](https://spark.brighte.com.au/package/fieldset#legend).

PR #128, opened 2022-05-24
Fix collapsing gap on Stack in Field

lukebennett88:
# Description

[Joss pointed out](https://brighte-au.slack.com/archives/C03C4J5AUHY/p1653353190316019) that the gap property was collapsing in certain scenarios.
This is because the gap was being set to `undefined` if the `labelVisibility` was set to "hidden". We did this because the Box that wraps the label was still appearing in the DOM, and the gap prop was causing space to be reserved for this hidden element (forgetting that the gap was needed for the description and message to be spaced properly.

In my testing, it looks like we can wrap the label with a fragment if labelVisibility is hidden, as Flexbox is smart enough to not reserve space for the `VisuallyHidden` label (I tested this in Brave (Chromium), Firefox and Safari).

## Associated Tickets

- [SPRK-92](https://brighte.atlassian.net/browse/SPRK-92)

### DIFF
--- a/packages/field/src/Field.tsx
+++ b/packages/field/src/Field.tsx
@@ -6,7 +6,6 @@ import { Stack } from '@spark-web/stack';
 import { Text } from '@spark-web/text';
 import { useTheme } from '@spark-web/theme';
 import type { DataAttributeMap } from '@spark-web/utils/internal';
-import { buildDataAttributes } from '@spark-web/utils/internal';
 import type { ReactElement, ReactNode } from 'react';
 import { forwardRef, Fragment } from 'react';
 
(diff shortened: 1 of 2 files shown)


packages/field/src/Field.tsx
```diff
@@ -115,22 +114,27 @@ export const Field = forwardRef<HTMLDivElement, FieldProps>(
...
+              {children}
+            </Box>
+          );
+
     return (
       <FieldContextProvider value={fieldContext}>
-        <Stack
-          gap={labelVisibility === 'hidden' ? undefined : 'medium'}
-          ref={forwardedRef}
-          {...(data ? buildDataAttributes(data) : null)}
```
lukebennett88, line 123: We can just pass the data prop down to the underlying `<Box>` so there is no need to call this function here anymore.

packages/field/src/Field.tsx
```diff
@@ -115,22 +114,27 @@ export const Field = forwardRef<HTMLDivElement, FieldProps>(
       ),
     };
 
+    const LabelWrapper =
+      labelVisibility === 'hidden'
+        ? Fragment
+        : ({ children }: { children: ReactNode }) => (
```
jossmac, line 120: Move this out of the main component. Where possible (almost always) we shouldn't create components at runtime.

packages/field/src/Field.tsx
```diff
@@ -115,22 +114,27 @@ export const Field = forwardRef<HTMLDivElement, FieldProps>(
       ),
     };
 
+    const LabelWrapper =
```
jossmac, line 117: Move this out of the main component. Where possible (almost always) we should avoid creating components at runtime.

jossmac (changes requested):

jossmac (approved):
LGTM

Issue #15, opened 2022-04-13
RadioGroup: a11y concerns


jossmac:
The [RadioGroup](https://spark.brighte.com.au/package/radio#radiogroup) component employs `display: contents`, which unfortunately can [negatively affect the accessibility](https://bugs.webkit.org/show_bug.cgi?id=185679) of the element it's applied to in some browsers.

A radio group without a "caption" presents some issues too, consider [fieldset](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset) + [legend](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/legend).

szhang-brighte:
Looks like this bug is fixed: https://bugs.webkit.org/show_bug.cgi?id=185679

szhang-brighte:
Looks like the bug is fixed: [here](https://bugs.webkit.org/show_bug.cgi?id=185679#c15)

PR #130, opened 2022-05-24
RadioCard component

lukebennett88:
# Description

Adds a RadioCard component to `@spark-web/radio`.
As part of this, I've refactored the `RadioPrimitive` quite a bit and changed the `RadioGroup` to address [this issue](https://github.com/brighte-labs/spark-web/issues/15).

# Associated Tickets

- [SPRK-14](https://brighte.atlassian.net/browse/SPRK-14)

# Checklist:

### DIFF
--- a/packages/radio/package.json
+++ b/packages/radio/package.json
@@ -14,7 +14,6 @@
     "@spark-web/box": "^1.0.5",
     "@spark-web/control-label": "^1.0.5",
     "@spark-web/field": "^3.0.0",
-    "@spark-web/icon": "^1.1.3",
     "@spark-web/stack": "^1.0.5",
     "@spark-web/text": "^1.0.5",
     "@spark-web/theme": "^3.0.1"
--- a/packages/radio/package.json
+++ b/packages/radio/package.json
@@ -14,6 +14,7 @@
     "@spark-web/box": "^1.0.5",
     "@spark-web/control-label": "^1.0.5",
     "@spark-web/field": "^3.0.0",
+    "@spark-web/fieldset": "^1.1.0",
     "@spark-web/stack": "^1.0.5",
     "@spark-web/text": "^1.0.5",
     "@spark-web/theme": "^3.0.1"
--- a/packages/radio/src/radio-group.tsx
+++ b/packages/radio/src/radio-group.tsx
@@ -1,4 +1,6 @@
 import { FieldMessage, useFieldIds } from '@spark-web/field';
+import { Fieldset } from '@spark-web/fieldset';
+import { Stack } from '@spark-web/stack';
 
 import { RadioGroupContext } from './context';
 import type { RadioGroupProps } from './types';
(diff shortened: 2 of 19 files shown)


packages/radio/package.json
```diff
@@ -14,7 +14,6 @@
     "@spark-web/box": "^1.0.5",
     "@spark-web/control-label": "^1.0.5",
     "@spark-web/field": "^2.0.0",
-    "@spark-web/icon": "^1.1.3",
```
lukebennett88, line 17: I've refactored how we display the radio so no longer need an SVG icon for this

packages/radio/src/types.ts
```diff
@@ -58,3 +58,7 @@ export type RadioGroupProps<Value extends string> = {
   /** The value of the nested radios. */
   value: Value;
 };
+
+export type RadioCardProps = Omit<RadioProps, 'id'> & {
```
lukebennett88, line 62: Need to remove size prop

packages/radio/src/radio-group.tsx
```diff
@@ -17,10 +21,14 @@ export const RadioGroup = <Value extends string>({
     <RadioGroupContext.Provider
       value={{ ...context, 'aria-describedby': message && messageId }}
     >
-      <div id={radioGroupId} role="radiogroup" style={{ display: 'contents' }}>
-        {children}
-      </div>
-      {message && <FieldMessage tone={tone} id={messageId} message={message} />}
+      <Fieldset legend={legend} gap={gap}>
```
jossmac, line 24: Probably for another PR. I think this might be a bit of an anti-pattern i.e. "inheritance" VS "composition".

The `Fieldset` should implement `Stack` with some "gap", without exposing the prop. Letting consumers compose however appropriate for a given design e.g.

```tsx
<Fieldset legend="Default behaviour">
  ...
</Fieldset>

<Fieldset legend="Custom stack">
  <Stack gap="large">
    ...
  </Stack>
</Fieldset>

<Fieldset legend="Custom whatever">
  <Row gap="large">
    ...
  </Row>
</Fieldset>
```

packages/radio/src/radio-primitive.tsx
```diff
@@ -1,134 +1,173 @@
...
-    backgroundColor: theme.color.background.surface,
+  const focusRingStyles = useFocusRing({ always: true });
 
-    borderColor: theme.border.color.field,
-    borderRadius: theme.border.radius.full,
-    borderStyle: 'solid',
-    borderWidth: theme.border.width.large,
+  const outerSize = sizeToScaleKey[size];
+  // Maps to 6px for small, and 9px for medium
+  const innerSize = theme.sizing[outerSize] / 2.6666666667;
```
jossmac, line 55: 😢

packages/radio/src/radio-primitive.tsx
```diff
@@ -1,134 +1,173 @@
...
+          {...inputProps}
+          {...radioStyles}
+          aria-checked={inputProps.checked}
+          aria-disabled={inputProps.disabled}
+          ref={forwardedRef}
+          as="input"
+          type="radio"
+        />
+        {/* Used for styling of Radio Card */}
+        <span aria-hidden data-radio-border="true" />
```
jossmac, line 36: This is interesting. The style implementation seems fairly divergent, might be worth considering a separate component.

packages/radio/src/radio-primitive.tsx
lukebennett88: Haha, I hate this as well, but a lookup object seemed like overkill and I couldn't come up with a name I liked. 
I will take another run at it.

packages/radio/src/radio-group.tsx
lukebennett88: The `Fieldset` _needs_ a `gap` prop (or equivalent) to control the space between the `<legend>` and the `children`.

In the vast majority of cases, I'd wager that the `gap` prop on the `Fieldset` should be _the same_ as the `gap` prop on the nested `Stack`.

Whilst I'm a fan of composition over inheritance, we needed to provide the prop in both places so it made sense to build this in so you only have to pass it once (if for some reason a consumer needed to offer a different amount of space between the children they could wrap children with _another_ `Stack`).

Do you have any suggestions on a different way to tackle this?

lukebennett88:
Closing this PR and opening up a new one here:
https://github.com/brighte-labs/spark-web/pull/164

PR #131, opened 2022-05-24
Populate props doc data for components 

matildaPan:
# Description

This is PR is ONLY for populating the props data for all components, in content layer. 
Later, we will utilise these data for UI rendering and adding to search index. 

- the main logic is in docs/utils/generate-props.ts and docs/contentlayer.config.ts
- the rest of changes are about adding props type and splitting components into seperate files


matildaPan:
switch to react-docgen-typescript

PR #134, opened 2022-05-26
feat(combobox): Added group type to items

yaminideveloper:
# Description

Please include a summary of the change and which issue is fixed. Please also
include relevant motivation and context. List any dependencies that are required
for this change.

# Associated Tickets

- SPRK-

# Checklist:

### DIFF
--- a/packages/combobox/src/combobox.tsx
+++ b/packages/combobox/src/combobox.tsx
@@ -2,7 +2,7 @@ import { useFieldContext } from '@spark-web/field';
 import type { DataAttributeMap } from '@spark-web/utils/internal';
 import { buildDataAttributes } from '@spark-web/utils/internal';
 import { useEffect, useRef, useState } from 'react';
-import type { GetOptionLabel, GetOptionValue } from 'react-select';
+import type { GetOptionLabel, GetOptionValue, GroupBase } from 'react-select';
 import ReactSelect from 'react-select';
 
 import {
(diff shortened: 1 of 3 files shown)


lukebennett88:
Hi @yaminideveloper, thanks for the PR.
We use Changesets to manage versioning. You could you please add one 🙏
https://github.com/brighte-labs/spark-web/blob/main/CONTRIBUTING.md#version-management

nderkim:
@lukebennett88 is this something we want to add? If so, it requires some more work to override the group styles to match our designs. I don't think we have designs for groups yet either.

packages/combobox/src/combobox.tsx
```diff
@@ -29,7 +29,7 @@ export type ComboboxProps<Item = unknown> = {
   inputValue?: string;
 
   /** Array of items for the user to select from. */
-  items: Awaitable<Item[]>;
+  items: Awaitable<(Item | GroupBase<Item>)[]>;
```
clothoo, line 32: Don't we need to pass it to react-select?

nderkim: Not necessarily. Our API intentionally only exposes a subset of the many features supported by `react-select`. If we want to add grouped items to our list of supported features, we will also need to add our styles to support it.

lukebennett88:
Updated styles for Group headings:
PixelSnap 2022-05-26 at 16 59 13@2x

lukebennett88 (approved):

PR #135, opened 2022-05-26
Update Playroom and Storybook dev env vars

lukebennett88:
# Description

Previously, when running the docs in development mode, Storybook and Playroom links, would link to the deployed site to ensure that you could always get to them.
In practice this wasn't very useful when you were trying to test things locally. So I've updated the links to point to their local servers instead.


nderkim (approved):
Excellent 😘

PR #136, opened 2022-05-27
Extract props data for use in search

jesstelford:
Using `react-docgen-typescript` (with a minor patch), we're extracting props from all components and storing them in the contentlayer JSON.

Later, we extract those props for use in the search component:

Screen Shot 2022-05-27 at 3 22 08 pm

Later, we will also use the props data to replace the manually created markdown table of props in the REAME.md's (coming in a future PR).


matildaPan (approved):
REALLY AWESOME

PR #142, opened 2022-06-01
feat(spinner): make spinner accessible

clothoo:
# Description

Currently, in test we have no way to tell whether the app is in a loading state.
Make Spinner accessible so that we can wait for it to disappear.

# Associated Tickets

# Checklist:

- [x ] I've updated unit tests where needed
- [x ] I've updated the components README.md where needed
- [x ] I've added major version bumps for my breaking changes
- [x ] I've added changesets where needed
- [x ] All the components I've updated are reflected in the changelog

### DIFF
--- a/.changeset/thirty-eels-perform.md
+++ b/.changeset/thirty-eels-perform.md
@@ -1,0 +1,5 @@
+---
+'@spark-web/spinner': patch
+---
+
+Add role 'progressbar'
--- a/packages/spinner/package.json
+++ b/packages/spinner/package.json
@@ -10,6 +10,7 @@
   "dependencies": {
     "@babel/runtime": "^7.18.0",
     "@emotion/css": "^11.9.0",
+    "@spark-web/a11y": "^1.0.5",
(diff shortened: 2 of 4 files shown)


lukebennett88:
Hey @clothoo, thats for the PR. Sorry for the slow response getting back to you on this.
I'm not sure that `role="progressbar"` is appropriate here, as it's the role of the `progress` element.
Would a `VisuallyHidden` span with some descriptive text work for you?

I'm thinking something like this:
CleanShot 2022-06-02 at 16 59 40@2x

clothoo:
> Hey @clothoo, thats for the PR. Sorry for the slow response getting back to you on this. I'm not sure that `role="progressbar"` is appropriate here, as it's the role of the `progress` element. Would a `VisuallyHidden` span with some descriptive text work for you?
> 
> I'm thinking something like this: CleanShot 2022-06-02 at 16 59 40@2x

Good point. I've updated the PR.

lukebennett88 (approved):

PR #144, opened 2022-06-02
Centre `HiddenWhenLoading` container inside of `Button`

lukebennett88:
# Description

I noticed that small icons were slightly misaligned when inside of a button. This PR fixes it!

Before:
PixelSnap 2022-06-02 at 17 06 22@2x

After:
PixelSnap 2022-06-02 at 17 07 06@2x


JedWatson (approved):

matildaPan (approved):
awesome

PR #146, opened 2022-06-06
Add className and style back to layout primitives

lukebennett88:
# Description

We removed a bunch of instances of prop spreading in <https://github.com/brighte-labs/spark-web/pull/99>.
I'm a little unsure about removing `className` and `style` props as there are valid use cases for them at the moment (e.g. with the `Stack` component in the docs, [we set an arbitrary width using `style`](https://spark.brighte.com.au/package/text#overflow-strategy)).

Whilst it's not recommended to use these props if you can help it, sometimes they are the lesser of two evils!

### DIFF
--- a/.changeset/itchy-papayas-divide.md
+++ b/.changeset/itchy-papayas-divide.md
@@ -1,0 +1,6 @@
+---
+'@spark-web/columns': minor
+'@spark-web/row': minor
+---
+
+Add className and style back to layout primitives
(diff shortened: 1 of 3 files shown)


JedWatson (approved):
Controversial, but fair 👍

lukebennett88:
To add a bit more context, this is only for layout primitives (Stack, Row, Inline, Columns and Box).
For pretty much everything else we _should_ be able to safely remove the className, style and spreads.

PR #148, opened 2022-06-07
Make Field a peer dep to avoid context bug

lukebennett88:
# Description

Some of our users were getting a bug when they tried to use an input that was wrapped in the Field component.

image

This was because the inputs had a dependency on `@spark-web/field`, and the version of Field that the inputs were using was different to the version they had installed themselves.

Moving the dependency for Field on these packages into peer-dependencies should fix this.


ChristopherMitchell242:
Nice update!!

ChristopherMitchell242 (approved):

PR #152, opened 2022-06-13
Docs/sprk 72 add package version to docs

ChristopherMitchell242:
# Description

The idea here is to let consumers know what version of the components docs is published and what maturity status we consider the component to be. 

Screen Shot 2022-06-14 at 9 05 54 am

# Associated Tickets

- SPRK-72

# Checklist:


docs/pages/package/[slug].tsx
```diff
@@ -81,6 +90,26 @@ export default function Packages({
...
+  return (
+    <Alert
+      tone={componentMaturityStatus === 'Experimental' ? 'caution' : 'info'}
+      heading={
+        componentMaturityStatus === 'Experimental'
+          ? 'Experimental'
+          : 'Early Adoption'
+      }
+    >
+      This component is considered {componentMaturityStatus}. Reach out to the
```
Comment (lukebennett88) [inline docs/pages/package/[slug].tsx:107]: The Alert is rendering without our typography styles as the child isn't a string (because `componentMaturityStatus` is being added inline, it's converted into an array of strings).

There are two ways to fix this:

Convert to a string so that this
```tsx
This component is considered {componentMaturityStatus}. Reach out to the
Spark team to understand more about what this means.
```
would become this:
```tsx
{`This component is considered ${componentMaturityStatus}. Reach out to
the Spark team to understand more about what this means.`}
```

Wrap the text with a Text component instead:
```tsx
<Text>
  This component is considered {componentMaturityStatus}. Reach out to the
  Spark team to understand more about what this means.
</Text>
```

Since we're already importing `Text`, I'd probably go with the second option.

docs/contentlayer.config.ts
```diff
@@ -67,6 +67,16 @@ export const Package = defineDocumentType(() => ({
         ).name;
       },
     },
+    componentMaturityStatus: {
+      type: 'string',
+      resolve: async pkg => {
```
lukebennett88, line 72: I'm wondering if we should simplify this.
Instead of looking at the `package.json` we look in the frontmatter, and since we only have experimental and stable, we can just make it a flag

```tsx
// contentlayer.config.ts

export const Package = defineDocumentType(() => ({
  // ...omitted for brevity
  fields: {
    isExperimentalPackage: {
      type: 'boolean',
      description: 'Maturity status of the component',
      required: false,
    },
  },
  // ...omitted for brevity
}));
```

Then in `[slug}.tsx`:
```tsx
// pages/package/[slug].tsx

export const getStaticProps: GetStaticProps<{
  // ...omitted for brevity
  isExperimentalPackage: boolean;
}> = async ({ params }) => {
  const pkg = allPackages.find(p => p.slug === params!.slug);
  if (!pkg) {
    return {
      notFound: true,
    };
  }

  return {
    props: {
      // ...omitted for brevity
      isExperimentalPackage: Boolean(pkg.isExperimentalPackage),
    },
  };
};
```

docs/pages/package/[slug].tsx
Comment (ChristopherMitchell242) [reply docs/pages/package/[slug].tsx:107]: Ah, good call, thanks mate, yup, will get that fixed up asap!

docs/contentlayer.config.ts
ChristopherMitchell242: Ok cool. Yeah this does seem like a better approach, will need to look at this more closely.

jesstelford: I'm 50/50 on putting it in the frontmatter; it feels like important information that could get lost easily (is it purely a decorative / docs "status", or does it have further impacts?) One future scenario where having it in the `package.json` is good; Publishing to npm under an `-experimental` label - whatever script which does that would have an easier time getting the info from `package.json`. OTOH; the only use right now is decorative, so having it in the README makes sense.

Either way, I prefer `componentMaturityStatus` over `isExperimentalPackage` as it provides flexibility for future statuses with no downsides today.

docs/pages/package/[slug].tsx
```diff
@@ -158,21 +158,20 @@ export default function Packages({
 }
 
 function ComponentMaturity({
-  componentMaturityStatus,
+  isExperimentalPackage,
 }: {
-  componentMaturityStatus: string;
+  isExperimentalPackage: Boolean;
```
Comment (lukebennett88) [inline docs/pages/package/[slug].tsx:163]: typo, should be lowercase "b" on `boolean`

lukebennett88 (approved):
LGTM 🚢 it 👍

PR #158, opened 2022-06-15
Update input styles

lukebennett88:
# Description

This PR adds a box shadow to our field inputs, and refactors the `useInput` hook (now renamed to `useInputStyles`) to return a tuple so we can prevent some duplication.

The first item returned from the array is an object containing props that should be spread onto the underlying `Box` element that our inputs are created with, and the second object is a CSS object that should be passed to Emotion's `css` function.

## Associated Tickets

- [SPRK-27](https://brighte.atlassian.net/browse/SPRK-27)

### DIFF
--- a/packages/select/src/Select.tsx
+++ b/packages/select/src/Select.tsx
@@ -3,7 +3,9 @@ import type { BoxProps } from '@spark-web/box';
 import { Box } from '@spark-web/box';
 import { useFieldContext } from '@spark-web/field';
 import { ChevronDownIcon } from '@spark-web/icon';
+import type { UseInputStylesProps } from '@spark-web/text-input';
 import { InputContainer, useInputStyles } from '@spark-web/text-input';
+import { useTheme } from '@spark-web/theme';
 import type { SelectHTMLAttributes } from 'react';
 import { forwardRef, useCallback } from 'react';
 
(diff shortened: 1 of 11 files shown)


packages/select/src/Select.tsx
```diff
@@ -20,11 +18,11 @@ export type OptionsOrGroups = Array<Option | Group>;
 export type SelectProps = Pick<
   SelectHTMLAttributes<HTMLSelectElement>,
   'defaultValue' | 'name' | 'onBlur' | 'onChange' | 'required' | 'value'
-> & {
-  data?: DataAttributeMap;
-  options: OptionsOrGroups;
-  placeholder?: string;
-};
+> &
+  Pick<BoxProps, 'data'> & {
```
lukebennett88, line 22: This lets us remove the dependency on the utils package.

packages/select/src/Select.tsx
```diff
@@ -55,21 +56,12 @@ export const Select = forwardRef<HTMLSelectElement, SelectProps>(
 
     return (
       <InputContainer>
+        <Indicator />
```
lukebennett88, line 59: Moved the chevron down icon into a component to make things a little more readable

packages/select/src/Select.tsx
```diff
@@ -111,24 +96,19 @@ export const Select = forwardRef<HTMLSelectElement, SelectProps>(
...
-    ...inputStyles,
-    overflow: 'hidden', // fix for Safari to prevent unwanted scrolling of parent container to occur
-    textOverflow: 'ellipsis',
-
-    // Prevent text going underneath the chevron icon
-    paddingRight:
-      theme.sizing.xxsmall + // size of chevron icon
-      theme.spacing.medium * 2, // paddingX value
-
-    ':invalid': {
```
lukebennett88, line 130: This has been moved to `useInputStyles()`

PR #159, opened 2022-06-16
Add Stepper component

lukebennett88:
# Description

WIP

# Associated Tickets

- SPRK-

# Checklist:


packages/stepper/src/step.tsx
```diff
@@ -0,0 +1,145 @@
...
+          ) : (
+            children
+          )}
+        </DefaultTextPropsProvider>
+      </Stack>
+    </Box>
+  );
+}
+
+Step.__isStep__ = true;
```
lukebennett88, line 72: Static property used when resolving children to throw an error if a child of `Stepper` is not a `Step`

packages/stepper/src/step.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+export const CircleIcon = createIcon(
+  <circle cx={12} cy={12} r={9} />,
+  'CircleIcon'
+);
+
+const fromStatus = {
```
lukebennett88, line 79: Todo: come up with a better name for this lookup object

packages/stepper/src/step.tsx
```diff
@@ -0,0 +1,145 @@
...
+  status: keyof typeof fromStatus;
+  heading: string;
+  secondaryHeading?: string;
+};
+
+export function Step({
+  children,
+  heading,
+  secondaryHeading,
+  status = 'disabled',
```
lukebennett88, line 25: It would be nice to come up with a way to infer the step instead of having to provide it for _every_ step.
e.g. in a linear flow anything that is _before_ the "current" must be complete, and anything after it should be disabled.

packages/stepper/src/index.ts
```diff
@@ -0,0 +1,6 @@
+export { useActiveIndex, useSetActiveIndex, useStepIndex } from './context';
+export { Step, Stepper } from './stepper';
+
+// types
```
ChristopherMitchell242, line 4: nit: think we can remove the comment?

ChristopherMitchell242 (commented):
Do we need a changeset for a brand new component?

PR #165, opened 2022-06-22
Update npm scripts

lukebennett88:
# Description

This PR updates the npm scripts in the monorepo.

I've made the verb the first part of scripts (e.g. `yarn run build:docs` instead of `yarn run docs:build`) as this makes more sense to me, and it's what we're doing in core. I've also added `npm-run-all` so we can do things like run `npm-run-all --parallel check:*` to check all the scripts that start with `check` in parallel (I've also added a new `validate` script to do just this.

I've removed all the variations of the `clean` script, as I've found that I only ever want to clear everything at once (and running `yarn clean` is easier to remember) and added a `fresh` script to remove all temporary files and run yarn again.

`yarn typecheck` becomes `yarn check:types` so that the `validate` script will automatically run it.

I've also updated the scripts in the docs to use the same prefix scheme.


package.json
```diff
@@ -8,27 +8,28 @@
...
-    "format:write": "prettier --write .",
+    "build:docs": "yarn --cwd=docs build",
+    "build:packages": "preconstruct build",
+    "check:format": "yarn run prettier --check",
+    "check:lint": "yarn run eslint",
+    "check:packages": "preconstruct validate",
+    "check:types": "tsc --noEmit && cd docs && yarn run check:types",
+    "clean": "git clean -fXd",
+    "dev:docs": "yarn --cwd=docs dev:docs",
+    "docs:build": "yarn run build:docs",
```
lukebennett88, line 19: This is just so the builds do fail. Once this is merged I'll change it in Vercel and add a PR to remove it.

clothoo (approved):

PR #171, opened 2022-07-05
feat(combobox): expose combobox hook

clothoo:
# Description

`useAwaitableItems` is useful in our hook form wrapper. Instead of duplicate the code, it'd be good to just expose it.

# Checklist:


.changeset/sweet-avocados-stare.md
```diff
@@ -0,0 +1,5 @@
+---
+'@spark-web/combobox': patch
```
lukebennett88, line 2: We're adding new functionality (letting consumers use this hooks) so it should be a minor release.

lukebennett88 (approved):

PR #174, opened 2022-07-12
Fix alignment when heading is omitted

lukebennett88:
# Description

When the heading prop is omitted, but you have a dismiss button the alignment was all off.
We were also using a hand-rolled button instead of the button component, and some of the colours were wrong (not bright enough — see secondary button in screenshot below).

Before:
PixelSnap 2022-07-13 at 09 01 53@2x

After:
PixelSnap 2022-07-13 at 09 02 04@2x

### DIFF
--- a/packages/alert/src/Alert.tsx
+++ b/packages/alert/src/Alert.tsx
@@ -1,4 +1,5 @@
-import { Box } from '@spark-web/box';
+import { css } from '@emotion/css';
+import { Button } from '@spark-web/button';
 import type { IconProps } from '@spark-web/icon';
 import {
   CheckCircleIcon,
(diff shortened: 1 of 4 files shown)


packages/alert/src/Alert.tsx
```diff
@@ -26,83 +35,81 @@ export type AlertProps = {
   tone: AlertTones;
 };
 
-export const Alert = ({
+export function Alert({
```
lukebennett88, line 38: Changed the arrow functions to regular functions because we're using hoisting. I think Babel must handle this for us, but I prefer to less magic!

packages/alert/src/Alert.tsx
```diff
@@ -26,83 +35,81 @@ export type AlertProps = {
...
         gap="medium"
         padding="large"
+        paddingRight="none"
+        width="full"
+        style={{ minWidth: 0 }}
       >
-        <IndicatorContainer>
-          <Icon tone={tone} size={iconSize} />
-        </IndicatorContainer>
+        <IconWrapper>
```
lukebennett88, line 69: Removed the dependency of the unrelated package (TextList) and created it inline.

packages/alert/src/Alert.tsx
```diff
@@ -26,83 +35,81 @@ export type AlertProps = {
...
-            alignItems="center"
-            justifyContent="center"
-            height="medium"
-            width="medium"
-            // NOTE: hover/focus styles etc have not been implemented
-            // as we don't have the correct tokens yet and this will soon be replaced
-            // with a `Button` component that will handle this for us.
-          >
+        <Row padding="small" alignSelf="start">
+          <Button label={closeLabel} tone={tone} prominence="low">
```
lukebennett88, line 79: No more hand-rolled `<Box as="button" />` 😅

dilipt (approved):

PR #176, opened 2022-07-15
Tabs component

lukebennett88:
# Description

Initial release of the new `Tabs` component.

Note: I'm not offering a controlled API just yet (that will come in the next PR). The API is also very strict about how the components are composed (e.g. `TabList` or `TabPanels` can not wrap their children in Fragments right now — which we might want to offer in the future).
In the future we might also want to consider offering an API to that lets us use links instead of buttons [like they do in Balance](https://balance.reckon.com/package/tabs#usage-as-navigational-elements).

# Associated Tickets

- [SPRK-57](https://brighte.atlassian.net/browse/SPRK-57)


package.json
```diff
@@ -55,11 +55,13 @@
     "@babel/preset-typescript": "^7.17.12",
     "@changesets/changelog-github": "^0.4.5",
     "@changesets/cli": "^2.23.0",
+    "@juggle/resize-observer": "^3.3.1",
```
lukebennett88, line 58: Needed some of these globals for test. Will create another PR adding them and rebase.

packages/tabs/src/tabs.tsx
```diff
@@ -0,0 +1,340 @@
...
+/**
+ * Tab
+ *
+ * The interactive element that changes the selected panel.
+ */
+
+const GAP = 'small'; // Space between the interactive element and the pseudo border
+
+export const Tab = forwardRef<HTMLButtonElement, TabProps>(
+  ({ children, data, disabled }, forwardedRef) => {
```
dilipt, line 115: I'm a little curious about the behaviour of a "disabled" tab. Can you tab to it with the keyboard etc. ? 
Is it visually distinct from an "enabled" one?

Or does this mean something else? 😅 Maybe it needs a storybook entry?

dilipt (approved):
🥇

packages/tabs/src/tabs.tsx
lukebennett88: There is actually an example in the docs:
https://spark-web-docs-git-feat-tabs-brighte.vercel.app/package/tabs#disabled-tabs

To answer your question though:
- No you can't navigate to it with your keyboard
- No it's not visually distinct (yet). Good callout and something we should absolutely address. None of our buttons are visually distinct when disabled, its actually the next thing I'm working on 😅

PR #178, opened 2022-07-20
Switch to using aria and data attributes for invalid styles

lukebennett88:
# Description

Our `useInputStyles` function (which is shared by a lot of our form components) was using the `:invalid` pseudo-class to style text, but our `Field` component uses inferred state from props for invalid styles so the styles didn't match up.

This PR changes the invalid styles to show up when `data-invalid` appears on the input.


ChristopherMitchell242 (approved):

packages/combobox/src/react-select-overrides.tsx
```diff
@@ -175,6 +181,11 @@ export const useReactSelectStylesOverride = <Item,>({
         },
       },
     }),
+
+    singleValue: provided => ({
+      ...provided,
+      '[data-invalid=true]': { color: theme.color.foreground.muted },
```
lukebennett88, line 187: Using a data attribute here as I'm pretty sure `aria-invalid` doesn't have any semantic meaning on a `<div>`.

PR #187, opened 2022-08-02
feat: add badge component

tamm:
# Description

Initial release of badge component.

A badge is a decorative indicator used to either call attention to an item or for communicating non-actionable, supplemental information.

# Checklist:

Screenshot of what this looks like now
Screen Shot 2022-08-02 at 14 02 28


tamm:
Need to map colours from status instead of background

packages/badge/src/Badge.tsx
```diff
@@ -0,0 +1,143 @@
...
+    <Box
+      data={data}
+      // Styles
+      display="inline-flex"
+      alignItems="center"
+      background="surface"
+      border="standard"
+      borderRadius="full"
+      gap="xsmall"
+      height="xsmall"
```
lukebennett88, line 30: @tamm I added a fixed height and moved the padding to be only on the x-axis which should address the issue you pointed out:

> The padding of 8px actually adds 1px too much, since in the design the height of the component is supposed to be 24px including the border, this would also need to be adjusted for as part of the left-padding

lukebennett88 (approved):

PR #223, opened 2022-09-14
bug(sprk-104) fix storybook not loading assets on the correct baseUrl & upgrade to webpack 5 

rickyrattlesnake:
## Description

When running the storybook local webpack-dev-server it fails to fetch `index.html` from the basePath. 

image

`yarn start-storybook` builds storybook using webpack and puts assets under the `node_modules/.cache/storybook/public` directory. 

The existing webpack config forces all asset resolution on the webpack-devserver to be from the `/storybook/` path, which isn't where the `iframe.html` for example lives. 

```typescript
  webpackFinal(config, options) {
      config.output!.publicPath = '/storybook/';
      return config;
  },
```

We only need to control this asset resolution because we build (for prod) with an output path `storybook` to differentiate it from playroom. 

So, solution was to remove all the custom asset paths for the dev server and relu on the existing `build-storybook -o ./public/storybook` flag to take care of correct setting the output directory for production build. We're not using the storybook webpack server in production, we don't need to set the webpack configuration for this. 

This PR
- Upgrades to Webpack 5 (doesn't fix the bug but it improves build times). https://storybook.js.org/docs/react/builders/webpack.
- Removes the PREVIEW_URL setting on the manager head html file which forces iframe.html to be fetched from a subpath. 
- In production we have to set the previewUrl so the iframe.html will load from a `/storybook/` subdirectory. There's an undocumented feature I used https://github.com/storybookjs/storybook/blob/next/code/lib/builder-webpack5/src/preview/iframe-webpack.config.ts#L206 to only do this in production. 
- See also https://github.com/storybookjs/storybook/issues/1291 and https://github.com/storybookjs/storybook/blob/7035ea7389393da041985ebc491ee58dedb50d06/code/lib/cli/src/build.ts#L12 for undocumented cli options to override preview_url and force building the preview
- Makes sure that dev builds will output to the default location and webpack resolve from the default public path.
- Edit nextjs dev config to point to storybook default path, since playroom and storybook are served on different ports, this should still work.
- Snuck in code splitting for slightly faster storybook load times.

## Associated Tickets

- [SPRK-104](https://brighte.atlassian.net/browse/SPRK-104)

## Checklist


rickyrattlesnake:
🤔  Ah i think vercel app is broken by this 

https://spark-web-docs-264s4o2wk-brighte.vercel.app/storybook?path=/story/page-layout-tabs--default doesn't work

lukebennett88:
> 🤔 Ah i think vercel app is broken by this

Yep noticed this as well. I think you just need to change `NEXT_PUBLIC_STORYBOOK_URL` in `.env.production` to be `/storybook/index.html`.
[This link works.](https://spark-web-docs-264s4o2wk-brighte.vercel.app/storybook/index.html?path=/story/page-layout-tabs--default)

docs/.storybook/main.ts
```diff
@@ -1,4 +1,6 @@
-module.exports = {
+import type { StorybookConfig } from '@storybook/core-common';
```
lukebennett88, line 1: 😎 much better

docs/.storybook/main.ts
```diff
@@ -7,12 +9,13 @@ module.exports = {
...
-    return config;
+  core: {
+    builder: 'webpack5',
   },
-  managerWebpack: async config => {
-    config.output.publicPath = '/storybook/';
-    return config;
+  features: {
+    // Allows code-splitting for faster load times
+    storyStoreV7: true,
```
lukebennett88, line 17: 👏

lukebennett88 (approved):
🚀

PR #226, opened 2022-09-14
SPRK-63 Automate documentation of props

nderkim:
## Description

Add props docs

## Associated Tickets

- [SPRK-63](https://brighte.atlassian.net/browse/SPRK-63)

## Checklist


lukebennett88 (changes requested):
PR looks good to me. We just need to get this ticket done first so we don't loose data that was only written in markdown.
https://brighte.atlassian.net/browse/SPRK-105

lukebennett88:
# Notes

## Accordion

- `type` is missing description

## Alert

- `icon` is missing description

## Button

- `children` is missing

## FloatInput

- `onChange` and `value` props are missing

## Hidden

- link in description for `on` prop is just text? Can we render description field for all props as markdown?

## `NextUtils

- using old hard-coded props table (will this work for functions or only for components?)

## Misc

- Can we filter out `as`, `asElement` and `ref`?
- `children` that are `ReactNode` is verbose — can we just use `ReactNode`

## Not related but noticed as part of this refactor

- ButtonLink probably shouldn't use `forwardRefWithAs`
- CurrencyInput missing heading for props
- PasswordInput missing props table altogether

lukebennett88 (approved):

PR #227, opened 2022-09-15
Rewrite docs for Box component

lukebennett88:
## Description

This is the first part of an effort to improve our docs. This PR updates the docs for the Box component.

## Associated Tickets

- [SPRK-36](https://brighte.atlassian.net/browse/SPRK-36)


packages/box/README.md
```diff
@@ -4,36 +4,227 @@ storybookPath: page-layout-box--default
 isExperimentalPackage: false
 ---
 
-Box is the lowest-level component for binding theme-based styles to an
-individual element on the screen. Internally, all Spark components are made up
-of Boxes.
+The Box component is our lowest-level primitive. The goal of the Box component
+is to map props to our tokens so that consumers should be able to use these
+props for all of their styling needs. Ideally you shouldn't need to use the
```
rickyrattlesnake, line 9:
```suggestion
props for all of their styling needs. Ideally, you shouldn't need to use the
```

packages/box/README.md
```diff
@@ -4,36 +4,227 @@ storybookPath: page-layout-box--default
...
+</Stack>
+```
+
+Notice that the Text in the example above doesn't use the tone prop, the colour
+is worked out using the BackgroundProvider in Box.
+
+**Note:** this will only work if you use the `background` prop. If you try to
+style the background colour in any other way the Text component will not know
+what colour the background is and so cannot invert it's colour to make sure that
+it's contents are readable.
```
rickyrattlesnake, line 223:
```suggestion
what colour the background is and so cannot invert its colour to make sure that
its contents are readable.
```

rickyrattlesnake (approved):

PR #237, opened 2022-09-19
WIP: Revise colour tokens

lukebennett88:
## Description

WIP

## Associated Tickets

- SPRK-

## Checklist


docs/components/content/index.tsx
```diff
@@ -172,7 +172,7 @@ const HeadingItem = ({
   const isActive = useIsActive(id);
 
   const textStyles = useText({
-    tone: 'neutral',
+    tone: 'neutralBold',
```
lukebennett88, line 175: I think we might want to do some mappings to remove `Bold`, `None` etc from the end of foreground tokens and let `useForegroundTone` work out the token to use based on the background.

PR #260, opened 2022-10-28
feat(combobox): Expose menuPortalTarget prop from react-select

szhang-brighte:
## Description

Combobox options are attached to `document.body` by default. Portal body css has a rule which makes the options smaller.

```css
body {
  font-size: 13px;
}
```
Instead of modifying Portal css (which is really messy), we use `menuPortalTarget` prop to render options inside a holding element for the options which uses the default font size (16px in most browsers). 
Screenshot 2022-10-28 at 20 44 36

## Associated Tickets

- SPRK-

## Checklist


yaminideveloper (approved):

ChristopherMitchell242 (commented):
Nice work

PR #263, opened 2022-11-08
Switch assert package

lukebennett88:
assert package uses nodejs util v12.5.0 which breaks vendor-microfrontends build.
This PR replaces the assert function with one from the [emery](https://emery-ts.vercel.app/) library instead.


szhang-brighte:
Looks good. Let me try it first. Thanks @lukebennett88

szhang-brighte (approved):

PR #266, opened 2022-11-09
fix: included name prop

yaminideveloper:
## Description
Included name props for floatInput so as to make it easy for testing

Please include a summary of the change and which issue is fixed. Please also
include relevant motivation and context. List any dependencies that are required
for this change.

## Associated Tickets

- SPRK-

## Checklist


lukebennett88:
Good catch! Looks like the required prop is also not getting forwarded.

I think a good idea here might be to spread the remaining props from everything based on the `TextInput` component (`DateInput`, `FloatInput`, `PasswordInput` etc) and to use React Aria's [filterDOMProps utility](https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/utils/src/filterDOMProps.ts) on the `TextInput` to ensure extra props still reach the underlying input unless they're not valid DOM attributes. That way, if you need do something we haven't accounted for, you can still use `@ts-ignore` to pass the attribute until the official fix makes its way back to the design system.

yaminideveloper:
> 

Thanks, @lukebennett88. `DateInput` and `PasswordInput` already spread all props of input, so I am just updating `FloatInput`. We have exposed valid props in TextInput, do we still need `filterDOMProps` utility?
image

packages/float-input/src/float-input.tsx
```diff
@@ -50,6 +51,7 @@ export const FloatInput = forwardRef<HTMLInputElement, FloatInputProps>(
         onChange={floatOnChange}
         onFocus={floatOnFocus}
         value={floatValue}
+        {...inputProps}
```
szhang-brighte, line 54: We need to put this line on line#45, otherwise the built-in props will be overridden by the spread props

szhang-brighte (approved):

lukebennett88:
@yaminideveloper the `filterDOMProps` isn't necessary as part of this PR (since we were already spreading consumer props onto the `TextInput` component, it would just be an enhancement to make things a little safer.

PR #293, opened 2023-02-05
SPAR-168: add background prop to accordion component

dilipt-brighte:
## Description

Allow for different background colours on the accordion component, which has popped up in a few designs.
Reduced the overall spacing and size of chevron as well.

## Associated Tickets

- SPAR-168

## Checklist


christopher-mitchell242 (approved):
Nice dude. 

Forget how the docs worked, I remember for a while having to manually update the docs for each component. But I do remember also trying it make them dynamic. 

I guess, we should make sure the docs are updated one way or another!

rozen-brighte (approved):
Nice!

